PR #4, opened 2021-02-16
Added configurable batch size option

usemam:
Added configurable batch size option to control how many work items are locked at once


cgillum (approved):
Thanks for this contribution!! I'll include it in the 0.5.0-alpha release.

Issue #18, opened 2021-04-16
Azure Managed Identity support


cgillum:
Currently we support connecting to a SQL database using Windows Integrated Auth or via username/password. Support for Azure Managed Identity would be helpful for users that are running in Azure (w/out Windows Integrated Auth support) and don't want to put database credentials in their application configuration.

Useful resource for instructions on how this works generally: https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/tutorial-windows-vm-access-sql

Useful code that can be borrowed for obtaining an access token: https://github.com/Azure/azure-functions-durable-extension/blob/47247f9ab322efdf9cd9216582bc5a7dce9bb37f/src/WebJobs.Extensions.DurableTask/ManagedIdentityTokenSource.cs#L55-L79

usemam:
@cgillum
Hi!
Could you please take a look at [this commit](https://github.com/usemam/durabletask-mssql/commit/732aec4a8600056aae15785eac0647971e09dbad)?
If it's not really bad - I'll create a PR.
Thanks!

cgillum:
@usemam this looks great! There are many a few tiny tweaks I would suggest, but if you're willing to create a PR, I can provide the more detailed feedback and we can get it merged!

PR #25, opened 2021-05-23
Azure Managed Identity support - first version

usemam:
These changes are to resolve issue #18

### DIFF
--- a/src/DurableTask.SqlServer.AzureFunctions/SqlDurabilityOptions.cs
+++ b/src/DurableTask.SqlServer.AzureFunctions/SqlDurabilityOptions.cs
@@ -24,6 +24,8 @@ namespace DurableTask.SqlServer.AzureFunctions
         [JsonProperty("taskEventBatchSize")]
         public int TaskEventBatchSize { get; set; } = 10;
 
+        public ManagedIdentityOptions? ManagedIdentityOptions { get; set; }
+
         internal ILoggerFactory LoggerFactory { get; set; } = NullLoggerFactory.Instance;
         
         internal SqlOrchestrationServiceSettings GetOrchestrationServiceSettings(
@@ -67,6 +69,16 @@ namespace DurableTask.SqlServer.AzureFunctions
             if (extensionOptions.MaxConcurrentOrchestratorFunctions.HasValue)
             {
                 settings.MaxActiveOrchestrations = extensionOptions.MaxConcurrentOrchestratorFunctions.Value;
+            }
+
+            if (this.ManagedIdentityOptions != null)
+            {
+                settings.ManagedIdentitySettings = new ManagedIdentitySettings
+                {
+                    UseAzureManagedIdentity = this.ManagedIdentityOptions.UseAzureManagedIdentity,
+                    AuthorityHost = this.ManagedIdentityOptions.AuthorityHost,
+                    TenantId = this.ManagedIdentityOptions.TenantId
+                };
             }
 
             return settings;
--- a/src/DurableTask.SqlServer/ManagedIdentitySettings.cs
+++ b/src/DurableTask.SqlServer/ManagedIdentitySettings.cs
@@ -1,0 +1,45 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System;
+using Newtonsoft.Json;
+
+namespace DurableTask.SqlServer
+{
+    /// <summary>
+    /// Configuration options for Managed Identity.
+    /// </summary>
+    public class ManagedIdentitySettings
+    {
+        public const string Resource = "https://database.windows.net/";
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ManagedIdentitySettings"/> class.
+        /// </summary>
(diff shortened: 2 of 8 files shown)


src/DurableTask.SqlServer.AzureFunctions/SqlDurabilityOptions.cs
```diff
@@ -24,6 +24,8 @@ class SqlDurabilityOptions
         [JsonProperty("taskEventBatchSize")]
         public int TaskEventBatchSize { get; set; } = 10;
 
+        public ManagedIdentityOptions? ManagedIdentityOptions { get; set; }
```
cgillum, line 27: Let's add `[JsonProperty("managedIdentityOptions")]` to this new property.

src/DurableTask.SqlServer/SqlConnectionFactory.cs
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Threading.Tasks;
+
+using Azure.Core;
+using Azure.Identity;
+
+using Microsoft.Data.SqlClient;
```
cgillum, line 9: Just for stylistic consistency, it would be great if you could move these `using` statements inside the `namespace` block.

src/DurableTask.SqlServer/SqlConnectionFactory.cs
```diff
@@ -0,0 +1,47 @@
...
+                if (this.managedIdentitySettings.AuthorityHost != null)
+                {
+                    azureCredentialOptions.AuthorityHost = this.managedIdentitySettings.AuthorityHost;
+                }
+                if (!string.IsNullOrEmpty(this.managedIdentitySettings.TenantId))
+                {
+                    azureCredentialOptions.InteractiveBrowserTenantId = this.managedIdentitySettings.TenantId;
+                }
+                var azureCredential = new DefaultAzureCredential(azureCredentialOptions);
+                var requestContext = new TokenRequestContext(new string[] { ManagedIdentitySettings.Resource });
```
cgillum, line 39: Would it make sense to cache this `requestContext` object?

src/DurableTask.SqlServer/SqlConnectionFactory.cs
```diff
@@ -0,0 +1,47 @@
...
+        readonly string connectionString;
+        readonly ManagedIdentitySettings? managedIdentitySettings;
+
+        public SqlConnectionFactory(string connectionString, ManagedIdentitySettings? managedIdentitySettings = null)
+        {
+            this.connectionString = connectionString;
+            this.managedIdentitySettings = managedIdentitySettings;
+        }
+
+        public async Task<SqlConnection> CreateConnection()
```
cgillum, line 24: Can we rename this to `CreateConnectionAsync()`? It's not a big deal, but we try to follow this naming convention for methods that return tasks (which is the case now that we may fetch an Azure AD token).

cgillum (commented):
This is great! I'd love to be able to release this ASAP, so I'll test it out today. Just a few very minor comments, but I can try to make the suggested changes if your PR allows me to make edits to it.

src/DurableTask.SqlServer.AzureFunctions/SqlDurabilityOptions.cs
cgillum: Unfortunately the Azure Functions configuration system doesn't like our use of nested options here. We'll need to flatten it in order for it to be recognized. My suggestion is that we add the three properties directly to the `SqlDurabilityOptions` class and rename them such that they all have the same prefix. For example, `azureManagedIdentityEnabled`, `azureManagedIdentityAuthorityHost`, and `azureManagedIdentityTenantId`.

src/DurableTask.SqlServer/ManagedIdentitySettings.cs
```diff
@@ -0,0 +1,45 @@
...
+using Newtonsoft.Json;
+
+namespace DurableTask.SqlServer
+{
+    /// <summary>
+    /// Configuration options for Managed Identity.
+    /// </summary>
+    public class ManagedIdentitySettings
+    {
+        public const string Resource = "https://database.windows.net/";
```
cgillum, line 14: In my testing, I noticed that I needed to add ".default" to the end of this URL before I could successfully fetch a token from Azure AD. I think this might be a newer requirement in the more modern versions of Azure.Identity.
```suggestion
        public const string Resource = "https://database.windows.net/.default";
```

cgillum:
I'd like to take this change today, so I'll go ahead and merge this PR and then add my own changes as a separate PR.

src/DurableTask.SqlServer/SqlConnectionFactory.cs
usemam: Sure. I guess I forgot about this naming pattern :)

src/DurableTask.SqlServer.AzureFunctions/SqlDurabilityOptions.cs
usemam: Sure, let's do it. Thanks for letting me know.

Issue #32, opened 2021-06-14
Update Microsoft.Data.SqlClient dependency to v3.0.0


cgillum:
Release notes: https://github.com/dotnet/SqlClient/blob/main/release-notes/3.0/3.0.0.md

Notable changes:
* This version has built-in support for Azure managed identities and could potentially allow us to simplify our own support for this.
* This version has built-in automatic retry support for transient errors (in preview). This could potentially replace our customer implementation.

usemam:
@cgillum 
Please take a look at PR #33 - it essentially reverts changes that introduced Azure Managed Identity support and adds reference to Microsoft.Data.SqlClient package v3.0.0

cgillum:
PR merged! This will go out with the v0.10.0 release.

PR #33, opened 2021-06-19
#32 Microsoft.Data.SqlClient v3.0.0

usemam:


cgillum (commented):
Thanks for this PR. I see you've removed the managed identity settings. How does managed identity work with the new v3.0.0 dependency? Can you provide some sample code that shows configuring an app with the SqlOrchestrationService with support for managed identities?

usemam:
Based on [this PR](https://github.com/dotnet/SqlClient/pull/1010) I assumed that managed identity parameters used for authentication now go through connection string.
I do apologize, but I can't currently check this assumption unfortunately - using my personal Azure subscription it will require spinning up AD domain, and I'm not sure that I have the knowledge (or resource) to do that right now. I will look into possible work-around for this.

usemam:
@cgillum 
After some more [source code](https://github.com/dotnet/SqlClient/tree/main/src) exploration, I was able to test my assumption, using custom Azure AD domain, Ubuntu 20.04 VM and Azure SQL database. I used [this tutorial](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/tutorial-windows-vm-access-sql) to setup VM and database.

Here's the code I ran on VM:
```
using System;
using System.Data.Common;
using Microsoft.Data.SqlClient;

namespace test_prog
{
    class Program
    {
        static void Main(string[] args)
        {
            using (var connection = new SqlConnection("Data Source = durable-sql.database.windows.net; Initial Catalog = DurableDB; Authentication = ActiveDirectoryManagedIdentity"))
            {
                connection.Open();
                using (var cmd = new SqlCommand())
                {
                    cmd.CommandText = "create table test_table(id int);";
                    cmd.Connection = connection;
		    cmd.ExecuteNonQuery();
                }
                Console.WriteLine("Hello World!");
            }
        }
    }
}
```
Looks like (in the simplest case at least) things like AuthorityHost and TenantId are not required to be provided in connection string (and I'm not sure they can be right now).

cgillum:
Excellent! Thank you for verifying! I'll do a bit more verification as well and then merge this PR. I'm very happy that the SQL client team made it much easier to integrate with managed identities.

cgillum:
Had a chance to do my own testing and everything looks great. I specifically tested locally using `Authentication=Active Directory Default` which is designed to work both locally and in Azure.

I've updated this PR to include a couple items in CHANGELOG.md. Thanks again for this contribution!!

Issue #36, opened 2021-06-26
DurableClient.ListInstancesAsync() forgets to convert TimeFrom/TimeTo to UTC


scale-tone:
Issue is similar to [this one](https://github.com/Azure/azure-functions-durable-extension/issues/1689).

Typically it is the callee method's responsibility to convert DateTime parameters to whatever time zone it wants them to be. That's also how the default storage provider behaves.

MSSQL provider doesn't seem to be doing that. As a result, if you pass From/To values with Type=Local, the method treats it as UTC and produces incorrect results.

PR #38, opened 2021-06-28
#36 SqlOrchestrationService.GetManyOrchestrationsAsync - convert spro…

usemam:
Attempt on addressing #36
SqlOrchestrationService.GetManyOrchestrationsAsync - convert sproc datetime parameters to UTC

### DIFF
--- a/src/DurableTask.SqlServer/SqlUtils.cs
+++ b/src/DurableTask.SqlServer/SqlUtils.cs
@@ -431,7 +431,7 @@ namespace DurableTask.SqlServer
             }
         }
 
-        public static SqlDateTime ToSqlDateTime(this DateTime dateTime, SqlDateTime defaultValue)
+        public static SqlDateTime ToSqlUtcDateTime(this DateTime dateTime, SqlDateTime defaultValue)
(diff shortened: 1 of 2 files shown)


src/DurableTask.SqlServer/SqlUtils.cs
```diff
@@ -445,10 +445,13 @@ public static SqlDateTime ToSqlDateTime(this DateTime dateTime, SqlDateTime defa
             {
                 return SqlDateTime.MinValue;
             }
-            else
+            else if (dateTime.Kind == DateTimeKind.Utc)
             {
                 return dateTime;
             }
+            else {
```
cgillum, line 452: Let's put this `{` on its own line for code-style consistency.

cgillum (commented):
Thanks for this PR! One small nit. Also, I think we'll need to check to see whether the CI test failure is an actual failure or if it's flakiness.

usemam:
Looks like it was a flake after all. All green now.

cgillum (approved):
Great - thanks again for this PR!

PR #39, opened 2021-06-28
Fix purge history threshold check

Jaah:
The documentation for `PurgeOrchestrationHistoryAsync` says it should be purging instances that are older than the given threshold. However it seems that the stored procedure used to retrieve these is currently retrieving instances _newer_ than the given threshold.

This will PR fixes the `PurgeInstanceStateByTime` stored procedure to retrieve instances that are _older_ than the given threshold time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## v0.10.0-beta
+
+### Breaking changes
+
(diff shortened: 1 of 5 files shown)


cgillum (approved):
Looks like this case is missing test coverage - but yes, this fix appears to be correct. Thank you for the contribution!

cgillum:
Looks like we have some tests that assume the previous behavior that also need to be updated in order for the CI to pass.

cgillum:
I went ahead and added the appropriate test updates to your PR. I also credited you with this contribution in the CHANGELOG. Thanks again!

Issue #48, opened 2021-10-21
Automatically Create Database If Not Present


wsugarman:
This extension is very easy-to-use with a simple configuration change and package import! But unlike the default Azure Storage option, the SQL provider requires that you create the database ahead of time. This isn't very convenient, especially because there may be settings, like the collation, that may only be discovered in the docs. I have also been playing around with azure functions using docker-compose, and it would require less overhead to simply spool up my SQL and Azure Function containers without executing an additional script to create my DB.

If that sounds like something you'd like to include as part of this storage provider, I have created a pull request to demonstrate the start of such a change.

PR #49, opened 2021-10-21
Automatically Create Database if Not Present

wsugarman:
As per https://github.com/microsoft/durabletask-mssql/issues/48, this change allows the worker to create the database automatically if it doesn't exist. Because the worker must gracefully handle the possibility the DB hasn't been created, a few changes needed to be made:
- The initial connection to the SQL server must use another `Initial Catalog` or `Database` instead of the user-specified one. As such, the worker now initially connects to the system database `master`
- The lock for upgrading the schema was also previously obtained within the context of the user-specified `Database`, so now the lock is only obtained after the presence of the database has been verified.
  - Each worker now independently checks the existence of the database.
- New log events have been created for executing arbitrary SQL commands/statements and creating a database
- New log assertion APIs have been added to accommodate non-deterministic scenarios, such as the initial inspection of whether the database exists

### DIFF
--- a/src/DurableTask.SqlServer/SqlDbManager.cs
+++ b/src/DurableTask.SqlServer/SqlDbManager.cs
@@ -30,8 +30,28 @@ namespace DurableTask.SqlServer
 
         public async Task CreateOrUpgradeSchemaAsync(bool recreateIfExists)
         {
+            // Note that we may not be able to connect to the DB, let alone obtain the lock,
+            // if the database does not exist yet. So we obtain a connection to the 'master' database for now.
+            SqlConnection connection = this.settings.CreateConnection("master");
+            await connection.OpenAsync();
+
+            if (!await this.DoesDatabaseExistAsync(this.settings.DatabaseName, connection))
+            {
+                if (!this.settings.CreateDatabaseIfNotExists)
+                {
+                    throw new InvalidOperationException($"Database '{this.settings.DatabaseName}' does not exist.");
+                }
+
+                await this.CreateDatabaseAsync(this.settings.DatabaseName, connection);
+            }
+
             // Prevent other create or delete operations from executing at the same time.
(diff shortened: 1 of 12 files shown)


src/DurableTask.SqlServer/SqlDbManager.cs
```diff
@@ -171,6 +173,57 @@ async Task<DatabaseLock> AcquireDatabaseLockAsync()
...
+            // if the database does not exist yet. So we obtain a connection to the 'master' database for now.
+            SqlConnection connection = this.settings.CreateConnection("master");
+            await connection.OpenAsync();
+
+            if (!await this.DoesDatabaseExistAsync(this.settings.DatabaseName, connection))
+            {
+                await this.CreateDatabaseAsync(this.settings.DatabaseName, connection);
+            }
+
+#if NETSTANDARD2_0
```
wsugarman, line 193: It's kind of ugly to bifurcate like this lol. I can use the same non-async overload for both frameworks too

test/DurableTask.SqlServer.Tests/Integration/DatabaseManagement.cs
```diff
@@ -163,37 +186,63 @@ public void SchemaCreationIsSerializedAndIdempotent()
...
+                .Expect(
+                    // 1st
+                    LogAssert.AcquiredAppLock(statusCode: 0),
+                    LogAssert.SprocCompleted("dt._GetVersions"),
+                    LogAssert.ExecutedSqlScript("schema-0.2.0.sql"),
+                    LogAssert.ExecutedSqlScript("logic.sql"),
+                    LogAssert.ExecutedSqlScript("permissions.sql"),
+                    LogAssert.SprocCompleted("dt._UpdateVersion"),
+                    // 2nd
+                    LogAssert.AcquiredAppLock(),
```
wsugarman, line 219: For the other workers, I don't think we can guarantee their response codes other than success; there may or may not be contention for the lock.

PR #68, opened 2022-02-02
fix the build issue of the Sample dockerfile for upgrading to sdk 6.0

TsuyoshiUshio:
The current sample Dockerfile doesn't work since the  PerformanceTest app is updgrade to the sdk 6.0.


cgillum (approved):
Thanks for the fix!

Issue #77, opened 2022-04-16
Support Purge instance


shreyas-gopalakrishna:
- Update the SqlOrchestrationService class to Implement the IOrchestrationServicePurgeClient interface and map it to the existing [PurgeOrchestrationHistoryAsync](https://github.dev/microsoft/durabletask-mssql/blob/ea38f5231543100702a57a48296f803cb058b143/src/DurableTask.SqlServer/SqlOrchestrationService.cs#L556) method.

Reference: https://github.com/microsoft/durabletask-java/issues/10
Sample Implementation PR for AzureStorageOrchestractionService: https://github.com/Azure/durabletask/pull/704

PR #80, opened 2022-04-25
Addressing #77 - Support instance purging

usemam:
Updated DurableTask.Core package ref to v2.9.0 and implemented new IOrchestrationServicePurgeClient interface for SqlOrchestrationService using some already existing mechanisms.
@cgillum please review.
Thanks.


cgillum (approved):
Thanks for this excellent PR! The implementation looks both clean and correct to me.

PR #88, opened 2022-04-29
#71 Add support for multi-instance queries

usemam:
Another PR - this one to implement IOrchestrationServiceQueryClient interface.

### DIFF
--- a/src/DurableTask.SqlServer/Scripts/logic.sql
+++ b/src/DurableTask.SqlServer/Scripts/logic.sql
@@ -1030,7 +1030,8 @@ CREATE OR ALTER PROCEDURE dt._QueryManyOrchestrations
     @CreatedTimeFrom datetime2 = NULL,
     @CreatedTimeTo datetime2 = NULL,
     @RuntimeStatusFilter varchar(200) = NULL,
-    @InstanceIDPrefix varchar(100) = NULL
+    @InstanceIDPrefix varchar(100) = NULL,
+    @TaskHubFilter varchar(250) = NULL
 AS
 BEGIN
     DECLARE @TaskHub varchar(50) = dt.CurrentTaskHub()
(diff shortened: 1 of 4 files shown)


src/DurableTask.SqlServer/Scripts/logic.sql
```diff
@@ -1060,7 +1061,8 @@ BEGIN
     FROM
         Instances I
     WHERE
-        I.[TaskHub] = @TaskHub AND
+        (@TaskHubFilter IS NULL AND I.[TaskHub] = @TaskHub OR 
+            I.[TaskHub] IN (SELECT [value] FROM string_split(@TaskHubFilter, ','))) AND
```
cgillum, line 1065: We don't normally allow the caller to specify any task hub information to the stored procedures. This is intentional to create a degree of security isolation between task hubs, which might be owned by different tenants. The task hub is typically determined by the connection string, which helps enforce this isolation, particularly when the task hub name comes from the login (which is unique for each tenant).

Allowing callers to provide a task hub filter that overrides the task hub inferred from the connection could allow one tenant to query information belonging to another tenant. For this reason, I'm not comfortable supporting this change.

cgillum (changes requested):
Thanks for this contribution! The structure of the change looks good but I have a concern about the task hub filter. Could we perhaps leave that out of this change?  In fact, we should probably throw a `NotSupportedException` if a task hub filter is provided in the query parameters.

src/DurableTask.SqlServer/Scripts/logic.sql
usemam: Thank you for this remark. I wasn't sure myself this filtration is necessary, since there are no other examples of filtering by hub name which is not current one for caller.
Removed filtration logic, and throwing `NotSupportedException` now if filter is present.

cgillum (approved):
Thank you for contributing this change. It looks great to me!

Apologies about the flakey CI. I will re-run it a few times and then merge this PR once I can confirm that all is good.

PR #92, opened 2022-05-03
Remove Microsoft.SqlServer.SqlManagementObjects dependency

### LINKED ISSUE #91: Possible to remove Microsoft.SqlServer.SqlManagementObjects dependency?
Hello! I noticed today that my function app's size increased from 13 MB to 35 MB after adding a dependency on this project. Curious about why the size jumped so much, I discovered the main culprit is this project's dependency on Microsoft.SqlServer.SqlManagementObjects. Digging into it further, I discovered that this dependency is only used for [a small piece of code in SqlDbManager](https://github.com/microsoft/durabletask-mssql/blob/main/src/DurableTask.SqlServer/SqlDbManager.cs#L235) that runs a SQL script. That code has a link to a SO article that was the inspiration for adding this dependency to begin with. However, on the same SO article, there are also solutions that don't require this dependency (such as [this one](https://stackoverflow.com/a/45595190/1785793)).

I'm unsure how feasible the alternatives actually are (though making the change locally at least didn't increase the number of failing tests for me :p) so apologies if this is completely unhelpful, but wanted to at least bring it up in case it could be a quick win.

IGx89:
Closes #91

### DIFF
--- a/src/DurableTask.SqlServer/DurableTask.SqlServer.csproj
+++ b/src/DurableTask.SqlServer/DurableTask.SqlServer.csproj
@@ -25,7 +25,6 @@
     <PackageReference Include="Microsoft.Data.SqlClient" Version="3.0.0" />
     <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="3.1.*" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="3.1.*" />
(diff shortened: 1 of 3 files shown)


cgillum (approved):
These changes look great - thanks for this contribution!! I will take care of merging it and including it in the next release.

hsnsalhi:
could you please tag this commit and push the nuget package so that I can use these new functionalities ? thank you

PR #96, opened 2022-05-28
Support rewinding orchestrations

### LINKED ISSUE #66: Support Rewind
Hello,

It seems that this provider does not support Rewind ATM, is this plan for futur, even if it's a preview feature with default storage provider it's quite usefull

Regards

Greybird:
Should fix #66.

I tried to mimic as much as possible the table storage implementation.
However, I'm pretty sure there is lot of room for improvement, so please don't hesitate to tell me where, and I'll try to enhance the proposal.

Known limitation:
- SQL Server seems to have a 32 calls depths limit, so as the rewind stored procedure is recursive, rewinding failed activities or orchestrations above this depth will not work (it should however be possible to rewind orchestrations with higher depth first, then going up in the hierarchy).


cgillum:
Thanks for this contribution! I did a quick scan of the changes and I don't have any issues with your approach. In the fullness of time, we're thinking of changing how we implement rewind to be storage provider-agnostic, but I think this implementation should be good enough for a preview-quality feature until we get around to a more generic overhaul.

src/DurableTask.SqlServer/Scripts/logic.sql
```diff
@@ -478,6 +478,20 @@ BEGIN
 END
 GO
 
+CREATE OR ALTER PROCEDURE dt.RewindInstance
+    @InstanceID varchar(100),
+    @Reason varchar(max) = NULL
+AS
+BEGIN
+    DECLARE @TaskHub varchar(50) = dt.CurrentTaskHub()
```
cgillum, line 486: Not used?

src/DurableTask.SqlServer/Scripts/logic.sql
```diff
@@ -1252,3 +1266,138 @@ BEGIN
...
+    FROM History h
+    WHERE h.[TaskHub] = @TaskHub AND h.[InstanceID] = @InstanceID
+      AND (h.[EventType] IN ('TaskFailed', 'SubOrchestrationInstanceFailed')
+           OR (h.[EventType] IN ('TaskScheduled', 'SubOrchestrationInstanceStarted') AND EXISTS (SELECT 1
+                                                           FROM History f
+                                                           WHERE f.[TaskHub] = @TaskHub AND f.[InstanceID] = @InstanceID  AND f.[TaskID] = h.[TaskID] AND f.[EventType] = CASE WHEN h.[EventType] = 'TaskScheduled' THEN 'TaskFailed' ELSE 'SubOrchestrationInstanceFailed' END)))
+
+    -- Mark all events related to failure as rewound
+    UPDATE Payloads
+    SET [Reason] = CONCAT('Rewound: ', ef.[EventType])
```
cgillum, line 1315: This seems a little strange. The `[Reason]` column is supposed to hold runtime data, but it looks like you're using it for rewind-specific metadata. I expect it will "just work" because none of the payload entries we're manipulating need a `[Reason]` value, but I may need to think about this a bit more.

src/DurableTask.SqlServer/Scripts/logic.sql
```diff
@@ -1252,3 +1266,138 @@ BEGIN
...
+    
+    -- List all events related to failures (ie TaskScheduled/TaskFailed and SubOrchestrationInstanceStarted/SubOrchestrationInstanceFailed couples)
+    INSERT INTO @eventsInFailure
+    SELECT h.[SequenceNumber], h.[EventType], h.[TaskID], h.[DataPayloadID]
+    FROM History h
+    WHERE h.[TaskHub] = @TaskHub AND h.[InstanceID] = @InstanceID
+      AND (h.[EventType] IN ('TaskFailed', 'SubOrchestrationInstanceFailed')
+           OR (h.[EventType] IN ('TaskScheduled', 'SubOrchestrationInstanceStarted') AND EXISTS (SELECT 1
+                                                           FROM History f
+                                                           WHERE f.[TaskHub] = @TaskHub AND f.[InstanceID] = @InstanceID  AND f.[TaskID] = h.[TaskID] AND f.[EventType] = CASE WHEN h.[EventType] = 'TaskScheduled' THEN 'TaskFailed' ELSE 'SubOrchestrationInstanceFailed' END)))
```
cgillum, line 1311: Making a note to myself to spend more time reviewing what this query is doing.

src/DurableTask.SqlServer/Scripts/logic.sql
```diff
@@ -1252,3 +1266,138 @@ BEGIN
...
+        WHERE i.[ParentInstanceID] = @InstanceID 
+          AND h.[EventType] = 'ExecutionStarted'
+          AND e.[EventType] = 'SubOrchestrationInstanceFailed'
+
+
+    OPEN subOrchestrationCursor 
+    FETCH NEXT FROM subOrchestrationCursor INTO @subOrchestrationInstanceID
+
+    WHILE @@FETCH_STATUS = 0 BEGIN
+        EXECUTE dt._RewindInstance @subOrchestrationInstanceID, @Reason
```
cgillum, line 1369: Should we add a comment here about the recursive depth limit?

src/DurableTask.SqlServer/Scripts/logic.sql
```diff
@@ -478,6 +478,20 @@ BEGIN
 END
 GO
 
+CREATE OR ALTER PROCEDURE dt.RewindInstance
```
cgillum, line 481: If it's not strictly necessary, I'd prefer we remove `dt.RewindInstance` and remove `dt._RewindInstance`. This is mainly because I want to avoid having people take a dependency on this sproc directly. In the future, we may change the implementation to be backend-agnostic, in which case we'd need to deprecate this "public" stored procedure.

cgillum (commented):
Some initial thoughts:

src/DurableTask.SqlServer/Scripts/logic.sql
Greybird: Good catch, should have removed this line!

src/DurableTask.SqlServer/Scripts/logic.sql
Greybird: Just as an initial help, it retrieves all `TaskFailed` and `SubOrchestrationInstanceFailed`, and their corresponding `TaskScheduled` and `SubOrchestrationInstanceStarted`.

src/DurableTask.SqlServer/Scripts/logic.sql
Greybird: I'm actually wondering if we could get rid of the recursion by collecting all instances with failures in the dependency tree at once. Do you think it is worth checking the feasibility ? Or maybe in a second change afterwards ?

src/DurableTask.SqlServer/Scripts/logic.sql
Greybird: I agree with you, I found it a little bit strange.
I chose to replicate as closely as possible the behaviour of Azure Table Storage found [here](https://github.com/Azure/durabletask/blob/12e4c5afc2e6463fd26cfb5f994c180685d45d15/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs#L357-L358), [here](https://github.com/Azure/durabletask/blob/12e4c5afc2e6463fd26cfb5f994c180685d45d15/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs#L384), and [here](https://github.com/Azure/durabletask/blob/12e4c5afc2e6463fd26cfb5f994c180685d45d15/src/DurableTask.AzureStorage/Tracking/AzureTableTrackingStore.cs#L393-L394), which is basically to keep track of the previous `EventType`, once replaced by `GenericEvent`.

Should `[Reason]` field be made part of the event table, or, if its `varchar(max)` nature has performance impacts, maybe split into its own table ?

Please tell me what direction you would like to take once you've had time to think about it!

cgillum: I think I may have missed this, but are these records being changed to GenericEvent and also having their Reason payloads set? I see that we’re doing this for Azure Storage. If so, then I have no problem with what you’re doing here. I haven’t reviewed very closely yet and thought maybe you were changing values for other event types.

src/DurableTask.SqlServer/Scripts/logic.sql
cgillum: Oops, I had a typo in my comment. I meant to suggest we keep just the private sproc. I’m fixing the comment to reflect this.

src/DurableTask.SqlServer/Scripts/logic.sql
Greybird: Yes that's what this code is doing. It updates payloads for the events who happen to already have one, and then add payloads (for the sole purpose of setting the reason) for those who haven't. Updated events are the ones that happen to have been switched to `GenericEvent`

src/DurableTask.SqlServer/Scripts/logic.sql
Greybird: Switched to two private proc (have to keep both as one is responsible for the transaction, which we do not want to reopen recursively.

Greybird:
I think I took care of your current comments.
Feel free to tell me if more is needed.

It's a great thing that Rewind is being worked on, to me it is a key feature of durable functions! Even better if it can be done in a storage-agnostic way!

src/DurableTask.SqlServer/Scripts/logic.sql
cgillum: Got it - thanks for clarifying this point. I didn't realize this is why you had two sprocs.

src/DurableTask.SqlServer/Scripts/logic.sql
cgillum: Thanks. I also took the liberty of reformatting this a bit to reduce horizontal scrolling, making it easier for me to read. :)

cgillum:
In an effort to make some small changes to your PR, it seems I've accidentally also merged in the latest from the main branch. Sorry about the noise it's causing on this PR. I'm basically good with these changes and will just do a few manual tests before signing off and merging.

Issue #86, opened 2022-04-28
Flakey CI test: MultiInstancePurge


cgillum:
This test fails very frequently in the CI, sometimes requiring multiple re-runs. Unfortunately, I'm not able to get it to fail locally, making it hard to root cause the issue.

PR #98, opened 2022-06-01
Use same precision for queries than for storage

Greybird:
I think the underlying issue of #86 was that the provider is issuing queries against the database using `datetime` parameters on `datetime2` fields, leading to a loss of precision in the queries for both purge and queries scenarii.  As sql server `datetime` is rounding at 0, 0.003, 0.007 seconds, while `datetime2` has a precision of 100ns (matching c# `DateTime` one), this leads to tests failing sometimes, but IMHO, it is also an actual issue in the provider code (even if probably a very small one).

I also made sure a delay of 20ms was left at each point in time capture in tests to ensure they are reliable (to compare with 100ns precision of `DateTime`/`datetime2`, ability of the OS to wait for such a low level of time accurately, and 1s which was previously used)

I was able to reproduce the #86 issue in around 20 seconds of running the `MultiInstancePurge` & `MultiInstanceQueries` test in a loop on my machine. With this PR code, no failing test has been reported after 2 hours.

### DIFF
--- a/src/DurableTask.SqlServer/SqlOrchestrationService.cs
+++ b/src/DurableTask.SqlServer/SqlOrchestrationService.cs
@@ -780,8 +780,8 @@ namespace DurableTask.SqlServer
                 throw new ArgumentOutOfRangeException(nameof(query), $"{nameof(query.PageNumber)} must be between 0 and {short.MaxValue} (inclusive).");
             }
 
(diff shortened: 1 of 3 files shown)


cgillum (approved):
Thank you for this! It was driving me nuts. I didn't realize that the problem was related to precision and `SqlDateTime`. I guess I'll know to stay away from this type from now on.

PR #99, opened 2022-06-01
add FetchParentInstancesOnly filter to fetch only parent instances if…

hsnsalhi:
Hello,
Refering to my issue at this link https://github.com/Azure/durabletask/issues/733
I created a Pull Request to filter results and retrieve only parent instances
This feature is required for our team and hope it'll be helpful for others.
I will create an other Pull Request in [durabletask](https://github.com/Azure/durabletask) to add the property FetchParentInstancesOnly  to OrchestrationQuery.
Thanks

### DIFF
--- a/src/DurableTask.SqlServer/Scripts/logic.sql
+++ b/src/DurableTask.SqlServer/Scripts/logic.sql
@@ -1059,7 +1059,8 @@ CREATE OR ALTER PROCEDURE dt._QueryManyOrchestrations
     @CreatedTimeFrom datetime2 = NULL,
     @CreatedTimeTo datetime2 = NULL,
     @RuntimeStatusFilter varchar(200) = NULL,
-    @InstanceIDPrefix varchar(100) = NULL
+    @InstanceIDPrefix varchar(100) = NULL,
+	@FetchParentInstancesOnly bit = 0
 AS
 BEGIN
     DECLARE @TaskHub varchar(50) = dt.CurrentTaskHub()
@@ -1093,7 +1094,8 @@ BEGIN
         (@CreatedTimeFrom IS NULL OR I.[CreatedTime] >= @CreatedTimeFrom) AND
         (@CreatedTimeTo IS NULL OR I.[CreatedTime] <= @CreatedTimeTo) AND
         (@RuntimeStatusFilter IS NULL OR I.[RuntimeStatus] IN (SELECT [value] FROM string_split(@RuntimeStatusFilter, ','))) AND
-        (@InstanceIDPrefix IS NULL OR I.[InstanceID] LIKE @InstanceIDPrefix + '%')
+        (@InstanceIDPrefix IS NULL OR I.[InstanceID] LIKE @InstanceIDPrefix + '%') AND
+		(@FetchParentInstancesOnly = 0 OR I.ParentInstanceID IS NULL)
     ORDER BY
         I.[CreatedTime] OFFSET (@PageNumber * @PageSize) ROWS FETCH NEXT @PageSize ROWS ONLY
 END
--- a/src/DurableTask.SqlServer/Scripts/logic.sql
+++ b/src/DurableTask.SqlServer/Scripts/logic.sql
@@ -1095,7 +1095,7 @@ BEGIN
         (@CreatedTimeTo IS NULL OR I.[CreatedTime] <= @CreatedTimeTo) AND
         (@RuntimeStatusFilter IS NULL OR I.[RuntimeStatus] IN (SELECT [value] FROM string_split(@RuntimeStatusFilter, ','))) AND
         (@InstanceIDPrefix IS NULL OR I.[InstanceID] LIKE @InstanceIDPrefix + '%') AND
-		(@ExcludeSubOrchestrations = 0 OR I.ParentInstanceID IS NULL)
+        (@ExcludeSubOrchestrations = 0 OR I.ParentInstanceID IS NULL)
     ORDER BY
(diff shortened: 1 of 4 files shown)


cgillum:
I would be fine with this PR if it could be done without requiring any changes to the DurableTask.Core dependency. Otherwise, we will require that this can be implemented for other major storage providers, particularly DurableTask.AzureStorage.

hsnsalhi:
Hello Chris and thank you for your reply 
I agree with you but the filter class is defined in DurableTask, this is why I created the PR for that repo (https://github.com/Azure/durabletask/pull/740) to add this property and maybe contribute on the other providers to implement the same feature.

PatriceOngla:
Hello Chris,
I understand your concern but this functionality is really useful and even necessary for a large number of scenarios. Maybe a good trade-off could be to implement it first for the sql server provider only and to clearly state in the documentation that the functionality is not yet available with the Azure Storage provider. We can implement it later on this clean basis.
What do you think ?

cgillum:
@PatriceOngla I'm fine with starting with just the SQL provider and then making it work more generically for other providers later. That would be the most expedient way to get this feature out. If you can update this PR to not rely on changes to DurableTask.Core and to also add some integration tests, then I can go ahead and accept this PR and include it in the next release.

src/DurableTask.SqlServer/SqlOrchestrationQuery.cs
```diff
@@ -53,5 +53,10 @@ public class SqlOrchestrationQuery
         /// Gets or sets an instance ID prefix to use for filtering orchestration instances.
         /// </summary>
         public string? InstanceIdPrefix { get; set; }
+
+        /// <summary>
+        /// Determines whether the query will retrieve only parent instances.
+        /// </summary>
+        public bool FetchParentInstancesOnly { get; set; } = false;
```
PatriceOngla, line 60: Naming: I suggest `FetchRootInstancesOnly`.

PatriceOngla:
Hello Chris,
Thank you for your answer. I can't really figure out how to implement this feature pretty much cleanly without modifying DurableTask.Core (the OrchestrationQuery class). Why not just throw a NotSupportedException for providers that do not implement this feature yet? This is already done for some other features (for example the TaskHubName criterion which triggers a NotSupportedException for the SqlServer provider if I remember well).

cgillum:
A SQL-only backend change would probably be to the SqlOrchestrationService.GetManyOrchestrationsAsync method. That shouldn't require any changes to DurableTask.Core.

DurableTask.AzureStorage accounts for >99.9% of usage right now. I don't want to add abstractions to DurableTask.Core which aren't supported by the mainstream implementation. I wouldn't mind if it were the other way around, i.e., new abstractions that are initially only supported by DurableTask.AzureStorage. In this case, however, I think it makes much more sense to start with this SQL backend since these kinds of queries will be _significantly_ more efficient on a relational database that can support multiple indexes. This kind of filter will be incredibly slow and expensive to implement on Azure Storage.

src/DurableTask.SqlServer/Scripts/logic.sql
```diff
@@ -1059,7 +1059,8 @@ CREATE OR ALTER PROCEDURE dt._QueryManyOrchestrations
     @CreatedTimeFrom datetime2 = NULL,
     @CreatedTimeTo datetime2 = NULL,
     @RuntimeStatusFilter varchar(200) = NULL,
-    @InstanceIDPrefix varchar(100) = NULL
+    @InstanceIDPrefix varchar(100) = NULL,
+	@FetchParentInstancesOnly bit = 0
```
cgillum, line 1063: nit: please replace tabs with spaces

src/DurableTask.SqlServer/Scripts/logic.sql
```diff
@@ -1093,7 +1094,8 @@ BEGIN
         (@CreatedTimeFrom IS NULL OR I.[CreatedTime] >= @CreatedTimeFrom) AND
         (@CreatedTimeTo IS NULL OR I.[CreatedTime] <= @CreatedTimeTo) AND
         (@RuntimeStatusFilter IS NULL OR I.[RuntimeStatus] IN (SELECT [value] FROM string_split(@RuntimeStatusFilter, ','))) AND
-        (@InstanceIDPrefix IS NULL OR I.[InstanceID] LIKE @InstanceIDPrefix + '%')
+        (@InstanceIDPrefix IS NULL OR I.[InstanceID] LIKE @InstanceIDPrefix + '%') AND
+		(@FetchParentInstancesOnly = 0 OR I.ParentInstanceID IS NULL)
```
cgillum, line 1098: nit: please replace tabs with spaces

src/DurableTask.SqlServer/SqlOrchestrationQuery.cs
cgillum: +1, or maybe `ExcludeSubOrchestrations`, which I like because it's just a verb and a noun. I also think something like `ExcludeSubOrchestrations` more clearly explains what this setting is doing.

The main problem with `FetchParentInstancesOnly` is that sub-orchestrations can also be parents if they create sub-orchestrations of their own, so the current name isn't actually completely accurate based on my reading of the code.

cgillum (commented):
Thanks! A few comments:

hsnsalhi:
For the failing test, I don't think it has anything to do with my PR, does it ?

cgillum (approved):
LGTM! Thanks for this contribution! I'll include it in the next release.

Issue #105, opened 2022-06-23
Feature Request - Custom Schema Name


moldovangeorge:
Currently, the multi-tenancy feature of this provider allows multiple services that share the same DB to use the same DT schema via TaskHub work isolation. 
In some scenarios, using the same schema across multiple services that are independently deployed will not work in the long term due to security and reliability concerns. 
We want to contribute to the provider by adding custom schema name support in a backward-compatible way but want to agree that this will be accepted by the DTF owners before starting any work on this.
For us, isolating each micro-service in their own DB in order to isolate the DTF work is not an option, and this contribution would highly improve our current DTF setup. 

Tagging @cgillum as a DTFx owner.

cgillum:
This is a feature I considered but avoided due to potential complexity and wanting to get community feedback first. Would you be willing to submit a design proposal for this? 

If we can make the problem statement clear and come up with a clean solution that doesn’t make the provider harder to use, then we can absolutely accept it as a contribution.

moldovangeorge:
Great, will come up with a proposal next week and will start from there. 👍

moldovangeorge:
Hei @cgillum, @AndreiRR24 worked on implementing my proposal and has everything ready on a feature branch to submit a PR. Can you give him write access to the repository so he can push the feature branch and open a PR?
We can start discussing the approach and the implementation once the PR is open.

cgillum:
I assume the feature branch is on a fork of this repo? If so, are you not able to open the PR from the fork? This is the normal process for submitting PRs and shouldn’t require write access.

mercer:
Yes, I can confirm the flow with the fork works https://github.com/microsoft/durabletask-mssql/pull/109

cgillum:
Closing as completed. Thank you so much for this valuable contribution! Note that I made some minor changes, which you can find here: https://github.com/microsoft/durabletask-mssql/pull/122. This new feature will be included in the v1.1.0 release and will be available for both DTFx users and Azure Functions users.

PR #110, opened 2022-07-24
Added multi-schema support for DTF Sql-Server

AndreiRR24:
Added multi-schema support for DTF Sql-Server as requested in #105.  Following the plan suggested by George, I've refactored the code to be able to use a custom schema name instead of the hardcoded "dt" as such:

-  Added a new property in SqlOrchestrationServiceSettings called SchemaName, defaulting to "dt"
- Modified all the scripts under /Scripts by replacing "dt" with a placeholder called {{SchemaNamePlaceholder}} and also added explicit table accessing by specifying the schemaName for each access.
- Added a new parameter to all xxxParameter methods from the SqlTypes files under /SqlTypes and from SqlUtils, called schemaName to be able to remove the hardcoded "dt".
- Modified the SqlDbManager such that before executing the script commands, it will replace the schemaNamePlaceholder with the schema name from the settings. I found it to be the easiest way to replace the "dt" from the script files without rewriting the whole script. There is still room for improvement on this part and any suggestion is highly appreciated.
- Added a new test file to test if the program is able to create and delete a schema with a custom name.
- Modified the test files such that it will be able to test custom table names by adding a new parameter to the test method and replacing all the occurrences of "dt" with said parameter value.

As such, this version is backwards compatible as far as I know and is able to do all the operations with custom schema names.

Co-Authored-By: George Moldovan <george.moldovan@uipath.com>

### DIFF
--- a/docs/multitenancy.md
+++ b/docs/multitenancy.md
@@ -8,7 +8,24 @@ One of the goals for the Microsoft SQL provider for the Durable Task Framework (
 
 Multitenancy works by isolating each app into a separate [task hub](taskhubs.md). The current task hub is determined by the credentials used to log into the database. For example, if your app connects to a Microsoft SQL database using **dbo** credentials (the default, built-in admin user for most databases), then the name of the connected task hub will be "dbo". Task hubs provide data isolation, ensuring that two users in the same database will not be able to access each other's data.
 
+Another layer of multitenancy is added by the multi-schema support. This increases reliability and security for multiple services that are independently deployed in the same database, allowing each service control over their own schema and further isolation between service's data.
+
 ?> Task hub isolation in the current version of the SQL provider prevents one tenant from accessing data that belongs to another tenant. However, it doesn't impose any restrictions on data volumes or database CPU usage. If this kind of strict resource isolation is required, then each tenant should instead be separated into their own database.
+
+## Managing custom schemas
+
+For self-hosted DTFx app that opt for custom schema name, you can configure the schema name directly in the `SqlOrchestrationServiceSettings` class.
+
+```csharp
+var settings = new SqlOrchestrationServiceSettings
+{
+    SchemaName = "customSchemaName",
+    TaskHubConnectionString = Environment.GetEnvironmentVariable("SQLDB_Connection"),
+};
+```
+
+If no schema name name is explicitly configured, the default value `dt` will be used. Note that changing the value requires a restart of the app for the change to take effect.
+
 
 ## Enabling multitenancy
 
--- a/src/DurableTask.SqlServer/SqlOrchestrationService.cs
+++ b/src/DurableTask.SqlServer/SqlOrchestrationService.cs
@@ -118,7 +118,7 @@ namespace DurableTask.SqlServer
             do
             {
                 using SqlConnection connection = await this.GetAndOpenConnectionAsync(cancellationToken);
-                using SqlCommand command = this.GetSprocCommand(connection, "dt._LockNextOrchestration");
+                using SqlCommand command = this.GetSprocCommand(connection, $"{this.settings.SchemaName}._LockNextOrchestration");
 
                 int batchSize = this.settings.WorkItemBatchSize;
                 DateTime lockExpiration = DateTime.UtcNow.Add(this.settings.WorkItemLockTimeout);
@@ -226,9 +226,9 @@ namespace DurableTask.SqlServer
                             // Delete the events and release the orchestration instance lock
                             using SqlCommand discardCommand = this.GetSprocCommand(
                                 connection,
-                                "dt._DiscardEventsAndUnlockInstance");
+                                $"{this.settings.SchemaName}._DiscardEventsAndUnlockInstance");
                             discardCommand.Parameters.Add("@InstanceID", SqlDbType.VarChar, 100).Value = instanceId;
-                            discardCommand.Parameters.AddMessageIdParameter("@DeletedEvents", messages);
+                            discardCommand.Parameters.AddMessageIdParameter("@DeletedEvents", messages, this.settings.SchemaName);
                             try
                             {
                                 await SqlUtils.ExecuteNonQueryAsync(
@@ -307,7 +307,7 @@ namespace DurableTask.SqlServer
         public override async Task RenewTaskOrchestrationWorkItemLockAsync(TaskOrchestrationWorkItem workItem)
         {
             using SqlConnection connection = await this.GetAndOpenConnectionAsync();
-            using SqlCommand command = this.GetSprocCommand(connection, "dt._RenewOrchestrationLocks");
+            using SqlCommand command = this.GetSprocCommand(connection, $"{this.settings.SchemaName}._RenewOrchestrationLocks");
 
             DateTime lockExpiration = DateTime.UtcNow.Add(this.settings.WorkItemLockTimeout);
 
@@ -334,7 +334,7 @@ namespace DurableTask.SqlServer
             Stopwatch sw = Stopwatch.StartNew();
 
             using SqlConnection connection = await this.GetAndOpenConnectionAsync();
-            using SqlCommand command = this.GetSprocCommand(connection, "dt._CheckpointOrchestration");
+            using SqlCommand command = this.GetSprocCommand(connection, $"{this.settings.SchemaName}._CheckpointOrchestration");
 
             OrchestrationInstance instance = newRuntimeState.OrchestrationInstance!;
             IList<HistoryEvent> newEvents = newRuntimeState.NewEvents;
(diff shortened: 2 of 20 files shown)


cgillum:
Awesome! Looking forward to going through this. I’m out of the country on vacation now but will do a full review when I get back in a few weeks.

src/DurableTask.SqlServer/SqlOrchestrationService.cs
```diff
@@ -118,7 +118,7 @@ public override Task DeleteAsync(bool deleteInstanceStore)
             do
             {
                 using SqlConnection connection = await this.GetAndOpenConnectionAsync(cancellationToken);
-                using SqlCommand command = this.GetSprocCommand(connection, "dt._LockNextOrchestration");
+                using SqlCommand command = this.GetSprocCommand(connection, $"{this.settings.SchemaName}._LockNextOrchestration");
```
keerthikiranvalathoru, line 122: Let's use a method for concatenation here, rather than every single line doing a string interpolation. We can also convert it to an extension method, so it can look like _LockNextOrchestration.PrependSchemaName()

src/DurableTask.SqlServer/SqlOrchestrationServiceSettings.cs
```diff
@@ -19,14 +19,15 @@ public class SqlOrchestrationServiceSettings
         /// </summary>
         /// <param name="connectionString">The connection string for connecting to the database.</param>
         /// <param name="taskHubName">Optional. The name of the task hub. If not specified, a default name will be used.</param>
-        public SqlOrchestrationServiceSettings(string connectionString, string? taskHubName = null)
+        public SqlOrchestrationServiceSettings(string connectionString, string? taskHubName = null, string? schemaName = null)
```
keerthikiranvalathoru, line 22: Nit: Please add another comment for new parameter

src/DurableTask.SqlServer/SqlTypes/MessageIdSqlType.cs
```diff
@@ -23,21 +21,23 @@ static class MessageIdSqlType
         public static SqlParameter AddMessageIdParameter(
             this SqlParameterCollection commandParameters,
             string paramName,
-            IEnumerable<TaskMessage> messageCollection)
+            IEnumerable<TaskMessage> messageCollection,
+            string schemaName)
         {
             SqlParameter param = commandParameters.Add(paramName, SqlDbType.Structured);
-            param.TypeName = SqlTypeName;
+            param.TypeName = $"{schemaName}.MessageIDs";
```
keerthikiranvalathoru, line 28: Looks like we are using a literal here instead of SqlTypeName

test/DurableTask.SqlServer.Tests/Integration/DatabaseManagement.cs
```diff
@@ -127,7 +128,61 @@ public async Task CanCreateAndDropSchema(bool isDatabaseMissing)
             // We know that all objects were successfully removed if the "dt" no longer exists.
             Assert.DoesNotContain("dt", testDb.GetSchemas());
         }
+        
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public async Task CanCreateAndDropSchemaWithCustomSchemaName(bool isDatabaseMissing)
+        {
```
keerthikiranvalathoru, line 136: We should have a test that checks no object gets created under dt schema, in case it is overriden by another name.

src/DurableTask.SqlServer/SqlOrchestrationService.cs
AndreiRR24: Added a new extension method to replace the interpolations and replaced all occurrences with the "PrependSchema()" method.

src/DurableTask.SqlServer/SqlTypes/MessageIdSqlType.cs
AndreiRR24: Replaced the literal with the extension method from the last suggestion.

test/DurableTask.SqlServer.Tests/Integration/DatabaseManagement.cs
AndreiRR24: On line 183 there is an assert checking there no "dt" schema and such, no objects under it. Is this what you were referring to?

test/DurableTask.SqlServer.Tests/Integration/DatabaseManagement.cs
```diff
@@ -127,7 +128,61 @@ public async Task CanCreateAndDropSchema(bool isDatabaseMissing)
...
+            LogAssert.NoWarningsOrErrors(this.logProvider);
+            LogAssert.Sequence(
+                this.logProvider,
+                LogAssert.AcquiredAppLock(),
+                LogAssert.ExecutedSqlScript("drop-schema.sql"));
+            
+            Assert.DoesNotContain($"{schemaName}", testDb.GetSchemas());
+            Assert.DoesNotContain("dt", testDb.GetSchemas());
+
+        }
```
cgillum, line 185: Can we add another test that adds multiple schemas to the same database? For example, the database is initially created with the "dt" schema and then we later set up an orchestration service using the "testSchemaName" schema. The end result should be two sets of tables. Even better if we can also verify that executing the drop scripts only impacts the target schema when multiple schemas are present.

cgillum (commented):
It's great that you were able to come up with a design for this that's non-breaking! I'm mostly happy with this change. The primary changes I'd like to see considered are 1) replacing the extension method with string interpolation (or a regular static method - I'd prefer that over an extension method in this case) and 2) another test that demonstrates deploying multiple schemas to the same database.

Also, would you be willing to make a pass through the documentation to help make it clear that the `dt` schema name can be customized?

AndreiRR24:
While writing the test @cgillum suggested, I encountered a problem with the LogAssert: After creating the first schema and validating the following steps, image the second schema validation fails with the following error 
image
even though after manual testing, the logic is right and the program is executing the flow correctly. This can be avoided by not running the second validation, but this doesn't sit quite right. Do you have any suggestions in solving this?

On another note, I do not know how to edit the documentation. I am guessing you mean [this one](https://microsoft.github.io/durabletask-mssql/#/) and not the README.md file.

cgillum:
> On another note, I do not know how to edit the documentation. I am guessing you mean [this one](https://microsoft.github.io/durabletask-mssql/#/) and not the README.md file.

Oh, sorry for not explaining this! Yes, I meant the published documentation. This documentation is authored via markdown files under the /docs directory. We're using GitHub Pages to have them published.

cgillum:
I'm not sure about the validation error. The log validation framework was created by another contributor @wsugarman, so I don't have as much insight into how it works internally to know why you're seeing these unexpected failures.

wsugarman:
> While writing the test @cgillum Chris Gillum FTE suggested, I encountered a problem with the LogAssert: After creating the first schema and validating the following steps, image the second schema validation fails with the following error image even though after manual testing, the logic is right and the program is executing the flow correctly. This can be avoided by not running the second validation, but this doesn't sit quite right. Do you have any suggestions in solving this?
> 
> On another note, I do not know how to edit the documentation. I am guessing you mean [this one](https://microsoft.github.io/durabletask-mssql/#/) and not the README.md file.

I haven't looked at this in awhile haha, but it seems that there is a missing log entry that was expected: `"secondTestSchemaName._UpdateVersion"`. In short, the tests record the log entries in a list, and the extensions validate the contents of the log through the `Expects` methods: https://github.com/microsoft/durabletask-mssql/blob/main/test/DurableTask.SqlServer.Tests/Logging/LogAssertExtensions.cs

src/DurableTask.SqlServer/SqlOrchestrationService.cs
AndreiRR24: Reverted the changes to use string interpolation again

test/DurableTask.SqlServer.Tests/Integration/DatabaseManagement.cs
AndreiRR24: Added a new test to verify that dropping one schema won't affect the other one and two sets of tables are created with the right names. Also added a few parameters to Get* functions to interact with the logger correctly.

docs/multitenancy.md
```diff
@@ -8,8 +8,25 @@ One of the goals for the Microsoft SQL provider for the Durable Task Framework (
...
 ?> Task hub isolation in the current version of the SQL provider prevents one tenant from accessing data that belongs to another tenant. However, it doesn't impose any restrictions on data volumes or database CPU usage. If this kind of strict resource isolation is required, then each tenant should instead be separated into their own database.
 
+## Managing custom schemas
+
+For self-hosted DTFx app that opt for custom schema name, you can configure the schema name directly in the `SqlOrchestrationServiceSettings` class.
+
+```csharp
+var settings = new SqlOrchestrationServiceSettings
+{
+    schemaName = "customSchemaName",
```
cgillum, line 22: This needs to be capitalized in order to compile:
```suggestion
    SchemaName = "customSchemaName",
```

src/DurableTask.SqlServer/SqlUtils.cs
```diff
@@ -26,7 +26,6 @@ static class SqlUtils
         {
             return reader.IsDBNull(columnIndex) ? null : reader.GetString(columnIndex);
         }
-
```
cgillum, line 29: nit: can we re-add this newline? I prefer to have whitespace between method declarations.

test/DurableTask.SqlServer.Tests/Integration/DatabaseManagement.cs
```diff
@@ -390,17 +539,20 @@ public IEnumerable<string> GetSchemas()
...
 
-            public IEnumerable<string> GetTables()
+            public IEnumerable<string> GetTables(string schemaName = "dt")
             {
                 this.testDb.Tables.Refresh();
                 foreach (Table table in this.testDb.Tables)
                 {
                     // e.g. "dt.History"
-                    yield return $"{table.Schema}.{table.Name}";
+                    if(table.Schema == schemaName)
```
cgillum, line 548: code style nit: please add a space after the `if` here, and on lines 566, 583, and 600.

cgillum (commented):
This is looking great, thank you! Just a few tiny cleanup things and then I think we can merge this PR!

AndreiRR24:
Updated the code with @cgillum's suggestions.

cgillum (approved):
This looks great, thank you again! I'll merge once the CI completes successfully.

EDIT: I noticed the validation workflow seems to be stuck in the "queued" state. I'm not sure if something is wrong with GitHub Actions right now, but I'll give it some time to resolve itself.

EDIT 2: I had to add a commit to get the CI to start again. Everything is green now!

Issue #120, opened 2022-08-18
CreateOrchestrationInstanceAsync contract not respected


mol-pensiondk:
De-dupe statuses passed to `CreateOrchestrationInstanceAsync` are not respected by the MSSQL backend. Also, the exception type thrown when an instance is not created due to duplicate ID is wrong.

In the following call, de-dupe statuses are explicitly given and include stauses beyond Pending and Running.
```
return await _taskHubClient.CreateOrchestrationInstanceAsync(
    "DurableTaskTest.TestOrchestration",
    null, // version
    oId,
    oInput,
    null, // tags
    new OrchestrationStatus[] {
        OrchestrationStatus.Running,
        OrchestrationStatus.Completed,
        OrchestrationStatus.ContinuedAsNew,
        OrchestrationStatus.Pending });
```
I use this to implement idempotency, allowing only restart of orchestrations which have failed. The Azure Storage backend works as expected and does not restart an already completed orchestration. The MSSQL back-end however does not respect the de-dupe parameter (indeed, it has a hardcoded list consisting of Pending and Running in the CreateInstance stored procedure).

Also, the exception thrown when a duplicate is detected is `InvalidOperationException`. I would have expected `DurableTask.Core.Exceptions.OrchestrationAlreadyExistsException`

cgillum:
Through code inspection, I confirmed that you're right on both issues.

One thing worth noting is that the Azure Storage backend also throws `InvalidOperationException` when a duplicate instance is created. I agree that changing this to `OrchestrationAlreadyExistsException` is the right thing to do. The good news is that `OrchestrationAlreadyExistsException` derives from `InvalidOperationException`, so it seems like we can make this change in a mostly non-breaking way.

mol-pensiondk:
Wow, that's fast you have a fix in the pipeline!

Regarding the Azure Storage backend throwning `InvalidOperationException`: Actually I can't get it to throw anything. It always returns an  `OrchestrationInstance`, regardless of whether the orchestration was created or not. It will have a non-existing `ExecutionId`, so is clearly the handle for the orchestartion that *would* have been created, if not for the de-dupe check.

PR #123, opened 2022-08-18
Fixes for deduple behavior on orchestration create

cgillum:
Fixes https://github.com/microsoft/durabletask-mssql/issues/120

Note that this change may be breaking, depending on how exception handling logic is written. I suspect in most, if not all cases, the change from `InvalidOperationException` to `OrchestrationAlreadyExistsException` should be non-breaking, hence, I'm releasing this fix as part of a minor version update instead of a major version update.

### DIFF
--- a/src/DurableTask.SqlServer/SqlOrchestrationService.cs
+++ b/src/DurableTask.SqlServer/SqlOrchestrationService.cs
@@ -14,6 +14,7 @@ namespace DurableTask.SqlServer
     using System.Threading.Tasks;
     using DurableTask.Core;
     using DurableTask.Core.Common;
+    using DurableTask.Core.Exceptions;
     using DurableTask.Core.History;
     using DurableTask.Core.Query;
     using DurableTask.SqlServer.SqlTypes;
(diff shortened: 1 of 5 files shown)


src/DurableTask.SqlServer/SqlOrchestrationService.cs
```diff
@@ -504,14 +505,19 @@ public override async Task CreateTaskOrchestrationAsync(TaskMessage creationMess
             command.Parameters.Add("@InputText", SqlDbType.VarChar).Value = startEvent.Input;
             command.Parameters.Add("@StartTime", SqlDbType.DateTime2).Value = startEvent.ScheduledStartTime;
 
+            if (dedupeStatuses?.Length > 0)
+            {
+                command.Parameters.Add("@DedupeStatuses", SqlDbType.VarChar).Value = string.Join(",", dedupeStatuses);
```
mol-pensiondk, line 510: What will happen if "Running" is not among the dedupeStatuses passed in? Will this clubber a running orchestration?

cgillum: Correct

mol-pensiondk: OK. Will not be a problem with our use, as "Running" is always included. I'm not sure what the Azure Storage backend would do?

mol-pensiondk: I just tested the behaviour of the Azure Storage backend. It seems that it will NOT clobber a running instance, even when "Running" is not among the dedupeStatuses. I don't get an exception (but then again, I never do with the Azure Storage backend) but the running orchestration will finish normally and not be restarted. Not so with "Pending" - a Pending orchestration _will_ be replaced with a new one, if Pending is not among the dedupeStatuses.

PR #128, opened 2022-09-17
Fix for #127 Is not properly using TaskHubName

### LINKED ISSUE #127: External client ignores task hub name configuration
Hello, 
When using an external client with sql provider. the specified task hub is ignored. The only way to have the specified task hub used is actually to use this
```CS
services.Configure<DurableTaskOptions>(options =>
        {
            options .HubName = "CustomTaskHub";
        });
```
But it's only one taskhub that could be use this way. And we need to be able to ask for many task hub on same sql schema.
What we do is 
```CS
private async Task<IReadOnlyCollection<DurableOrchestrationStatus>> SearchPerTaskHubAsync(string taskHub,
        OrchestrationStatusQueryCondition condition, CancellationToken cancellationToken)
    {
        var client = _durableClientFactory.CreateClient(new DurableClientOptions
        {
            TaskHub = taskHub // <-- this is not used
        });
        var result = new List<DurableOrchestrationStatus>();
        while (true)
        {
            var page = await client.ListInstancesAsync(condition, cancellationToken);
            result.AddRange(page.DurableOrchestrationState);
            if(string.IsNullOrEmpty(page.ContinuationToken))
                return result;
            condition.ContinuationToken = page.ContinuationToken;
        }
    }
```

bhugot:
Fixed #127 Durability factory when calling for a client is using default TestHubName or DurableTaskOption HubName instead of using the specified TaskHub


test/DurableTask.SqlServer.AzureFunctions.Tests/IntegrationTestBase.cs
```diff
@@ -152,11 +164,12 @@ protected async Task<DurableOrchestrationStatus> RewindOrchestrationAsync(string
             return status;
         }
 
-        protected async Task<IDurableClient> GetDurableClientAsync()
+        protected IDurableClient GetDurableClient(string? taskHubName = null)
         {
-            var clientRef = new IDurableClient[1];
-            await this.CallFunctionAsync(nameof(ClientFunctions.GetDurableClient), "clientRef", clientRef);
-            IDurableClient client = clientRef[0];
+            var client = this.clientFactory.CreateClient(new DurableClientOptions
```
cgillum, line 169: Nice to see that you found a way around my ugly hack for getting a client. For some reason I never considered using the client factory. 😅

test/DurableTask.SqlServer.AzureFunctions.Tests/IntegrationTestBase.cs
```diff
@@ -21,21 +23,31 @@ namespace DurableTask.SqlServer.AzureFunctions.Tests
 
     public class IntegrationTestBase : IAsyncLifetime
     {
+        private readonly string taskHubName;
```
cgillum, line 26: Please remove `private` for consistency with the rest of the code. This code base doesn't use default visibility modifiers.

test/DurableTask.SqlServer.AzureFunctions.Tests/CoreScenarios.cs
```diff
@@ -154,7 +154,155 @@ public async Task MultiInstanceQueries()
...
+            // Test ShowInput
+            condition.ShowInput = true;
+            condition.CreatedTimeFrom = sequencesFinishedTime;
+            result = await client.ListInstancesAsync(condition, CancellationToken.None);
+            Assert.All(result.DurableOrchestrationState, state => Assert.NotEqual(JValue.CreateNull(), state.Input));
+
+            condition.ShowInput = false;
+            condition.CreatedTimeFrom = startTime;
+            result = await client.ListInstancesAsync(condition, CancellationToken.None);
+            Assert.All(result.DurableOrchestrationState, state => Assert.Equal(JValue.CreateNull(), state.Input));
```
cgillum, line 268: Based on how the diff worked out, it looks like the core test logic is the same for both your new test and this older test. Is it possible to refactor this common code into a separate static method?

test/DurableTask.SqlServer.AzureFunctions.Tests/CoreScenarios.cs
```diff
@@ -154,7 +154,155 @@ public async Task MultiInstanceQueries()
...
+        }
+
+         [Fact]
+        public async Task MultiInstanceQueriesWithoutMultiTenancy()
+        {
+            await SharedTestHelpers.DisableMultitenancyAsync();
+            await SharedTestHelpers.PurgeAsync();
```
cgillum, line 275: Is the only difference between these two tests the disabling of multitenancy? Can you clarify what issues this variant catches that the existing test doesn't?

src/DurableTask.SqlServer.AzureFunctions/SqlDurabilityProviderFactory.cs
```diff
@@ -82,14 +82,13 @@ public DurabilityProvider GetDurabilityProvider(DurableClientAttribute attribute
                 }
 
                 SqlDurabilityOptions clientOptions = this.GetSqlOptions(attribute);
-                IOrchestrationServiceClient serviceClient = 
-                    new SqlOrchestrationService(clientOptions.GetOrchestrationServiceSettings(
+                SqlOrchestrationService orchestrationService = 
+                    new (clientOptions.GetOrchestrationServiceSettings(
```
cgillum, line 86: Should this instead be `SqlOrchestrationService orchestrationService = this.GetOrchestrationService()`?

cgillum (commented):
Thanks for this PR! I few questions/comments:

src/DurableTask.SqlServer.AzureFunctions/SqlDurabilityProviderFactory.cs
bhugot: Nope the problem come from this.GetOrchestrationService() that is not taking care of give options

test/DurableTask.SqlServer.AzureFunctions.Tests/CoreScenarios.cs
bhugot: I added 2 more clear tests on the difference between  with and without

test/DurableTask.SqlServer.AzureFunctions.Tests/CoreScenarios.cs
bhugot: I called the original test from the new one as it's more a regression check

test/DurableTask.SqlServer.AzureFunctions.Tests/MultiTenancyCoreScenarios.cs
```diff
@@ -0,0 +1,41 @@
+namespace DurableTask.SqlServer.AzureFunctions.Tests
```
cgillum, line 1: Please add the common copyright header to the top of this new file.

test/DurableTask.SqlServer.AzureFunctions.Tests/Functions.cs
```diff
@@ -0,0 +1,164 @@
+namespace DurableTask.SqlServer.AzureFunctions.Tests
```
cgillum, line 1: Please add the common copyright header to the top of this new file.

test/DurableTask.SqlServer.AzureFunctions.Tests/MultiTenancyCoreScenarios.cs
```diff
@@ -0,0 +1,41 @@
...
+
+            IDurableClient client = this.GetDurableClient();
+            var current = await client.GetStatusAsync(currentTaskHubInstanceId);
+            Assert.NotNull(current);
+            var otherInstance = await client.GetStatusAsync(anotherTaskHubInstanceId);
+            Assert.NotNull(otherInstance);
+
+            IDurableClient clientOtherTaskHub = this.GetDurableClient(otherTaskHubName);
+            var currentFromOtherTaskHub = await clientOtherTaskHub.GetStatusAsync(currentTaskHubInstanceId);
+            Assert.NotNull(currentFromOtherTaskHub);
```
cgillum, line 35: This is confusing to me. If you have two different task hubs, why should a client configured for one be able to fetch status for the other?

bhugot: It should not but it s the case and that what leaded to the bug for me. Maybe instead we should throw if a custom taskhub is used when in multitenancy mode

test/DurableTask.SqlServer.AzureFunctions.Tests/MultiTenancyCoreScenarios.cs
bhugot: Done

test/DurableTask.SqlServer.AzureFunctions.Tests/Functions.cs
bhugot: Done

test/DurableTask.SqlServer.AzureFunctions.Tests/MultiTenancyCoreScenarios.cs
bhugot: I removed it but I think there should be an issue about this.

bhugot:
Hello @cgillum any news on this?

test/DurableTask.SqlServer.AzureFunctions.Tests/IntegrationTestBase.cs
```diff
@@ -139,9 +165,22 @@ protected Task CallFunctionAsync(string name, IDictionary<string, object?>? args
             return status;
         }
 
+
+        protected async Task StartOrchestrationWithoutWaitingAsync(
```
cgillum, line 169: Why is this new method needed? I see it's used in the new test you added, but I don't understand why it's preferred over the existing `StartOrchestrationAsync`. If you don't strictly need the "WithoutWaiting" behavior, then I'd prefer we use the existing method, which provides better determinism.

test/DurableTask.SqlServer.AzureFunctions.Tests/MultiTenancyCoreScenarios.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace DurableTask.SqlServer.AzureFunctions.Tests
+{
+    using Xunit.Abstractions;
+
+    public class MultiTenancyCoreScenarios : CoreScenarios
```
cgillum, line 8: I'd rather not introduce an empty class. Can we instead make CoreScenarios a concrete class again but keep the WithoutMultiTenancyCoreScenarios.cs derivative?

cgillum (commented):
Thanks for the ping, and so sorry for dropping the ball on this PR. My only remaining ask is if we can drop a few things that I don't think are fully necessary. After that, I'm good to accept this.

test/DurableTask.SqlServer.AzureFunctions.Tests/IntegrationTestBase.cs
bhugot: The client WaitForStartAsync is failing due to the fact is waiting for another status than pending but as it's on another taskhubname the other one is never change from pending I will fix the first case

test/DurableTask.SqlServer.AzureFunctions.Tests/MultiTenancyCoreScenarios.cs
bhugot: Done

test/DurableTask.SqlServer.AzureFunctions.Tests/IntegrationTestBase.cs
bhugot: If you prefer another name I can change it

bhugot:
All good for me @cgillum

test/DurableTask.SqlServer.AzureFunctions.Tests/IntegrationTestBase.cs
cgillum: Why would it be on another task hub name? The client that starts the new orchestration is the same one that's waiting for it to complete.

bhugot: Yes the client but not the worker

cgillum: oh, it's because they are being started on a task hub that doesn't have a corresponding worker, right? I that case, this makes sense.

bhugot:
@cgillum I fixed the test it was timeout due to purge but it's not required in multitenancy tests

cgillum:
It looks like there might be a test concurrency issue now that we have multiple integration test classes . Here's the error from the CI:

```
Content root path: /home/runner/work/durabletask-mssql/durabletask-mssql/test/DurableTask.SqlServer.AzureFunctions.Tests/bin/Debug/net6.0/
  Failed DurableTask.SqlServer.AzureFunctions.Tests.WithoutMultiTenancyCoreScenarios.MultiInstanceQueries [1 ms]
  Error Message:
   System.Net.Sockets.SocketException : Address already in use
```

The solution mentioned [here](https://stackoverflow.com/a/47174842/2069) might be good to look into.

bhugot:
Thx for the tips @cgillum was looking at it. could you rerun I hope attribute on CoreScenario is going to work with inheritance

bhugot:
Great

cgillum (approved):
Alright, everything looks good. Thank you for this contribution!

PR #130, opened 2022-10-31
Added upgrade test infrastructure

### LINKED ISSUE #124: Database upgrade testing
Now that the first stable release is out, we need to ensure that subsequent releases are sufficiently backwards compatible. To help ensure backwards compatibility across releases, we should add upgrade tests. An example of what this might look like:

* Create a database using the v1.0.0 libraries
* Populate the v1.0.0 database with a mix of orchestrations, some completed, some running and waiting for events, and some pending (using scheduled start times).
* Backup the database and check it into the Git repository
* Create a test (or tests) in the vCurrent code that restore the v1.0.0 database, followed by;
  * Going through the normal app startup flow, which should auto-update the schema
  * Allow pending instances to start normally
  * Create new instances using the new instances
  * Resume existing instances
  * Query for all instances, old and new, asserting that all expected properties are available.

This should give us a good sense about whether any given PR is sufficiently backwards compatible with earlier database versions.

As we release new versions, we can add new tests that do the same thing, but starting from different previously released versions. For example, a v4 release could include tests that validate upgrades from v1 to v4, v2 to v4, and v3 to v4.

Ideally, we have a new upgrade test for every minor release (minor releases typically contain new features). New tests shouldn't be necessary for patch releases.

### LINKED ISSUE #191: App fails to start up when upgrading from 1.0.0 to 1.2.0 schema
This is a regression from this PR: https://github.com/microsoft/durabletask-mssql/pull/190. This change has not been released as a nuget package yet, so no existing users should be impacted unless they are building from source.

The error message is as follows:

> Microsoft.Data.SqlClient.SqlException : Cannot drop type 'dt.HistoryEvents' because it is being referenced by object '_CheckpointOrchestration'. There may be other objects that reference this type.

A fix is needed to ensure that custom types can be updated when they are referenced by stored procedures.

cgillum:
Resolves #124 
Resolves #191

This PR introduces a tool for generating SQL databases with data from the v1.0.0 release of the SQL provider as well as a new test that restores the generated database and runs tests against it. This is necessary to ensure that databases can be auto-upgraded and that no data loss occurs as part of the upgrade.


src/DurableTask.SqlServer/DurableTask.SqlServer.csproj
```diff
@@ -21,7 +21,7 @@
   </ItemGroup>
   
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.DurableTask.Core" Version="2.10.0" />
+    <PackageReference Include="Microsoft.Azure.DurableTask.Core" Version="2.11.1" />
```
davidmrdavid, line 24: Would it make sense for us to replace the version with `2.*`? My hope is that this would reduce the need to keep the version up to date.

test/DurableTask.SqlServer.Tests/Utils/SharedTestHelpers.cs
```diff
@@ -29,7 +29,7 @@ public static string GetTestName(ITestOutputHelper output)
         public static string GetDefaultConnectionString(string database = "DurableDB")
         {
             // The default for local development on a Windows OS
-            string defaultConnectionString = $"Server=localhost;Database={database};Trusted_Connection=True;";
+            string defaultConnectionString = $"Server=localhost;Database={database};Trusted_Connection=True;Encrypt=False;";
```
davidmrdavid, line 32: Can you clarify why we need to disable encryption?

test/DurableTask.SqlServer.Tests/DurableTask.SqlServer.Tests.csproj
```diff
@@ -21,4 +21,10 @@
     <ProjectReference Include="..\..\src\DurableTask.SqlServer\DurableTask.SqlServer.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <None Update="DatabaseBackups\DurableDB-v1.0.0.bak.zip">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+  </ItemGroup>
```
davidmrdavid, line 28: I noticed we committed this file. Was this intentional? I would assume that perhaps we only want this generated and cleaned up at test time.

I just worry about keeping `.zip` files in the commit history, which sometimes have weird diffs over time.

tools/TestDBGenerator/Program.cs
```diff
@@ -0,0 +1,110 @@
...
+    {
+        string? connectionString = Environment.GetEnvironmentVariable("SQLDB_Connection");
+        if (!string.IsNullOrEmpty(connectionString))
+        {
+            return connectionString;
+        }
+
+        Console.Error.WriteLine("Specify the database connection string in the 'SQLDB_Connection' environment variable.");
+        Environment.Exit(1);
+        return string.Empty;
```
davidmrdavid, line 108: nitpick: do we need the `return` statement here? It appears to me that `Environm.Exit()` would immediately terminate the program.

tools/TestDBGenerator/Program.cs
```diff
@@ -0,0 +1,110 @@
...
+{
+    static async Task Main(string[] args)
+    {
+        SqlConnectionStringBuilder builder = new(GetConnectionString());
+        Server dbServer = new(new ServerConnection(new SqlConnection(builder.ToString())));
+
+        // There's an assumption that the database schema version matches the assembly version.
+        Version assemblyVersion = typeof(SqlOrchestrationService).Assembly.GetName().Version!;
+        Version schemaVersion = new(assemblyVersion.Major, assemblyVersion.Minor, assemblyVersion.Build);
+
```
davidmrdavid, line 22: can you elaborate on why we have this assumption? Is this something we should be checking against in this test? Does anything automatically go wrong if this assumption is not met?

tools/TestDBGenerator/Orchestrations.cs
```diff
@@ -0,0 +1,211 @@
...
+using DurableTask.Core.History;
+using DurableTask.SqlServer;
+
+static class Orchestrations
+{
+    static readonly ManualResetEvent ActivityBlockedEvent = new(initialState: false);
+    
+    static bool PauseActivityExecution { get; set; } = false;
+
+    public static async Task GenerateRuntimeDataAsync(SqlOrchestrationService service)
```
davidmrdavid, line 19: this method needs to create a very specific end state for the workflows it schedules, which isn't immediately obvious to me unless I'm cross checking with the `UpgradeTests.cs` checks.

Any chance we could document these runtime end states in here? Would make this method;s purpose easier to understand.

test/DurableTask.SqlServer.Tests/Integration/UpgradeTests.cs
```diff
@@ -0,0 +1,306 @@
...
+            }
+
+            public override T Create()
+            {
+                return (T)Activator.CreateInstance(this.type)!;
+            }
+        }
+
+        // NOTE: Must be kept consistent with the version defined in tools/TestDBGenerator/Orchestrations.cs
+        class ActivitySequence : TaskOrchestration<string[], object>
```
davidmrdavid, line 246: if these need to be consistent, is it possible for us to have a single definition that is referenced across both files?

davidmrdavid (commented):
Left some thoughts and questions 🦖

src/DurableTask.SqlServer/DurableTask.SqlServer.csproj
cgillum: Hmm...I used to think this didn't work, but reading the [documentation](https://learn.microsoft.com/en-us/nuget/concepts/package-versioning#floating-version-resolutions) again, I don't see any reason why it wouldn't. This would be a good thing for us to revisit across all of our projects, including those AssemblyInfo.cs files that we use in .NET Isolated projects.

test/DurableTask.SqlServer.Tests/Utils/SharedTestHelpers.cs
cgillum: This should be the default behavior already for the version of the SQL SDK that we're using. Just making it explicit. In [the most recent version (4.0.0)](https://github.com/dotnet/SqlClient/blob/main/release-notes/4.0/4.0.0.md#breaking-changes), we'll be required to set this explicitly if we want our tests to work with arbitrary SQL container instances (like the one we use for the CI tests).

test/DurableTask.SqlServer.Tests/DurableTask.SqlServer.Tests.csproj
cgillum: Yes, this was intentional. I suppose we could generate it during the test, but that would require a lot more test code, not to mention other complexities once we do upgrade tests for multiple versions. These zip files should never change after being generated. The idea is that this backed up database (and any other versions we add here) should be _forever_ compatible with the SQL provider (or at least until we release a v2.x version).

tools/TestDBGenerator/Program.cs
cgillum: Yes, `return` is needed in order for the program to compile. It obviously should never run, but unlike `throw` statements, the C# compiler doesn't understand that no code will execute after  `Environment.Exit()`.

tools/TestDBGenerator/Program.cs
cgillum: This is the behavior of our DB upgrade logic, which you can find [here](https://github.com/microsoft/durabletask-mssql/blob/a0a6a5f9f75d8a1fa5501c107c154a021f2cd7d8/src/DurableTask.SqlServer/SqlDbManager.cs#L30-L121). If this assumption is broken, then the upgrade test should fail because the database schema won't be upgraded properly. That said, we can probably add an assert that validates this explicitly.

tools/TestDBGenerator/Orchestrations.cs
cgillum: Do the existing comments inside the method not clarify what the expected outcomes are? Maybe you can clarify what you mean by "end state"?

test/DurableTask.SqlServer.Tests/Integration/UpgradeTests.cs
cgillum: My reasoning for keeping separate copies of these is that one copy must never be changed (it needs to represent a snapshot of the type of code we supported in the v1.0.0 release of the SQL provider) whereas the other can be changed as appropriate based on future product updates we make.

davidmrdavid: I see. So essentially the one that should not change is `Orchestrations.cs` and the one that may be changed is the one in the tests?

tools/TestDBGenerator/Program.cs
davidmrdavid: Got it. That file actually helped give me a lot of context around how this is supposed to work.
I think the test sufficiently checks for correctness, so I won't ask for an explicit check of this assumption.

test/DurableTask.SqlServer.Tests/Integration/UpgradeTests.cs
cgillum: Right - it's fine to change the test code (provided it's semantically equivalent).

tools/TestDBGenerator/Orchestrations.cs
davidmrdavid: By end state I meant the checks that we will be making for these orchestrations. For example, here we say that "Test2" is scheduled with a delay, but that doesn't make it immediately obvious that the reason for this is that we want it to remain in the "Running" state until the DB is restored. It makes sense in retrospective, but when I first read through this file (and before I got to the tests) I wasn't sure what we were trying to accomplish with this data generation step. I think the comments in describe more explicitly the state of all these orchestrators when the taskhub-worker is shut down.

_But_ now that I've re-read the comments in this file, I find my feedback here to be too nitpicky so I'll retract it :) .
At most,, I would just suggest updating the comment here for "Test2" to make it clearer that we expect the shutdown worker to shut down before it gets a chance to run.

test/DurableTask.SqlServer.Tests/Integration/UpgradeTests.cs
davidmrdavid: In any case, could we document as a comment in code which file is meant to remain static? That wasn't immediately obvious to me and it seems calling out. @cgillum

src/DurableTask.SqlServer/DurableTask.SqlServer.csproj
davidmrdavid: @cgillum: nice! So do we want to introduce that change here or punt it for a separate PR?

test/DurableTask.SqlServer.Tests/Integration/UpgradeTests.cs
davidmrdavid: I think this is my only blocking suggestion at this point.

src/DurableTask.SqlServer/DurableTask.SqlServer.csproj
cgillum: I went ahead and introduced this change in another PR.

test/DurableTask.SqlServer.Tests/Integration/UpgradeTests.cs
cgillum: I've added comments at the top of TestDBGenerator/Orchestrations.cs, which should hopefully help clarify this.

tools/TestDBGenerator/Orchestrations.cs
cgillum: I added a bit more commentary as part of https://github.com/microsoft/durabletask-mssql/pull/130/commits/c164131f63c135b0dfa4121172234d5385eaf179, which will hopefully make the intent more clear.

src/DurableTask.SqlServer/Scripts/schema-1.2.0.sql
```diff
@@ -25,6 +25,33 @@ IF NOT EXISTS (SELECT 1 FROM sys.columns WHERE object_id = OBJECT_ID('__SchemaNa
...
+		     ON d.referencing_id = o.[object_id]
+	      INNER JOIN sys.schemas AS s
+		     ON o.[schema_id] = s.[schema_id]
+	      WHERE d.referenced_database_name IS NULL
+		    AND d.referenced_class_desc = 'TYPE'
+            AND d.referenced_entity_name IN ('HistoryEvents', 'OrchestrationEvents', 'TaskEvents')
+    ) [types]
+*/
+
+-- First, drop the rferencing stored procedures
```
bachuv, line 46: typo: "referencing"

bachuv (approved):
The schema-1.2.0.sql changes make sense to me. Thanks for adding the description that explains how to find the stored procedure dependencies!

PR #134, opened 2023-01-02
Synchronous reads for improved performance for large payloads

### LINKED ISSUE #126: Heavy Payload lead to stuck orchestration
Hello, we met a problem with sql provider as the payload is stored in table when he is too big it's truncated. and after that we have some timeout and nothing is processed. 
I think we should use the azure storage way to handle it by file in that case.

bhugot:
…on  because of mssql client async bug [sqlcli](https://github.com/dotnet/SqlClient/issues/593)

It should be easy to switch back when the fix is done
It should also fix #126

### DIFF
--- a/src/DurableTask.SqlServer/SqlUtils.cs
+++ b/src/DurableTask.SqlServer/SqlUtils.cs
@@ -398,7 +398,7 @@ namespace DurableTask.SqlServer
                 command,
                 traceHelper,
                 instanceId,
-                cmd => cmd.ExecuteReaderAsync(cancellationToken));
+                cmd => cmd.ExecuteReader());
         }
 
         public static Task<int> ExecuteNonQueryAsync(
@@ -432,6 +432,32 @@ namespace DurableTask.SqlServer
             LogHelper traceHelper,
             string? instanceId,
             Func<DbCommand, Task<T>> executor)
+        {
+            var context = new SprocExecutionContext();
+            try
+            {
+                return await WithRetry(() => executor(command), context, traceHelper, instanceId);
+            }
+            finally
+            {
+                context.LatencyStopwatch.Stop();
+                switch (command.CommandType)
+                {
+                    case CommandType.StoredProcedure:
+                        traceHelper.SprocCompleted(command.CommandText, context.LatencyStopwatch, context.RetryCount, instanceId);
+                        break;
+                    default:
+                        traceHelper.CommandCompleted(command.CommandText, context.LatencyStopwatch, context.RetryCount, instanceId);
+                        break;
+                }
+            }
+        }
+
+        static async Task<T> ExecuteSprocAndTraceAsync<T>(
+            DbCommand command,
+            LogHelper traceHelper,
+            string? instanceId,
+            Func<DbCommand, T> executor)
         {
             var context = new SprocExecutionContext();
             try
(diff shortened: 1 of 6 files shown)


src/DurableTask.SqlServer/SqlUtils.cs
```diff
@@ -398,7 +398,7 @@ static IEnumerable<SqlDataRecord> GetInstanceIdRecords(IEnumerable<string> insta
                 command,
                 traceHelper,
                 instanceId,
-                cmd => cmd.ExecuteReaderAsync(cancellationToken));
+                cmd => cmd.ExecuteReader());
```
cgillum, line 401: Instead of making this change for all calls to `ExecuteReaderAsync`, I would rather we make a separate `ExecuteReader` API and selectively choose which operations use sync vs. async. My preference is that `IOrchestrationService` interface methods use the synchronous variant whereas the `IOrchestrationServiceClient` methods continue to use the async variant. The reason for the latter is that an unknown number of threads may try to call `IOrchestrationServiceClient` methods, and I want to avoid the possibility of running into thread starvation for those calls.

src/DurableTask.SqlServer/SqlUtils.cs
```diff
@@ -453,6 +453,32 @@ static IEnumerable<SqlDataRecord> GetInstanceIdRecords(IEnumerable<string> insta
...
+                context.LatencyStopwatch.Stop();
+                switch (command.CommandType)
+                {
+                    case CommandType.StoredProcedure:
+                        traceHelper.SprocCompleted(command.CommandText, context.LatencyStopwatch, context.RetryCount, instanceId);
+                        break;
+                    default:
+                        traceHelper.CommandCompleted(command.CommandText, context.LatencyStopwatch, context.RetryCount, instanceId);
+                        break;
+                }
```
cgillum, line 478: Can we factor the code in this finally block into a shared `TraceSprocResult(...)` helper method?

src/DurableTask.SqlServer/SqlUtils.cs
```diff
@@ -529,6 +555,50 @@ static async Task<T> WithRetry<T>(Func<Task<T>> func, SprocExecutionContext cont
...
+                    // we could delay as long as 0 + 1 + 2 + 3 + 4 = 10 total seconds.
+                    TimeSpan delay = TimeSpan.FromSeconds(context.RetryCount);
+                    lock (random)
+                    {
+                        // Add a small amount of random delay to distribute concurrent retries 
+                        delay += TimeSpan.FromMilliseconds(random.Next(100));
+                    }
+
+                    // Log a warning so that these issues can be properly investigated
+                    traceHelper.TransientDatabaseFailure(e, instanceId, context.RetryCount);
```
cgillum, line 592: Can we factor this `catch` logic (except for the delay) in a shared helper method? Perhaps it can have a signature like `TimeSpan GetExceptionRetryDelayOrThrow(...)`.

cgillum (commented):
Thanks for this PR! I'd like to reduce its scope a little so that it only impacts execution-related SQL calls. I'd also like it if we can try to reduce some of the duplication.

bhugot:
@cgillum a new version is coming as my fix was incorrect the performance issue what on the reader and not on ExecuteReaderAsync I let a first run to show the test performance (on my computer it was around 1 min 40 and after the fix it was 12sec)

cgillum:
@bhugot I just verified your fix on my machine: your stress test takes 1.1 minutes without the fix and is 4.5 seconds with your fix. That's pretty incredible!

bhugot:
:)

cgillum (approved):
@bhugot thanks for this great PR! I added some minor changes to clean up a few things. Let me know if you have any questions or concerns about my changes.

bhugot:
It's ok for me

PR #136, opened 2023-01-05
Add dotnet-isolated extensions package

jviau:
This PR enables mssql usage in Azure functions dotnet isolated. 

- Adds a new project and nuget package `Microsoft.Azure.Functions.Worker.Extensions.DurableTask.SqlServer`, which has a single attribute that tells the host to load the `Microsoft.Azure.DurableTask.SqlServer.AzureFunctions` package.
    - Customers will reference this package in their dotnet isolated projects (and _not_ `Microsoft.Azure.DurableTask.SqlServer.AzureFunctions`)

### DIFF
--- a/src/Functions.Worker.Extensions.DurableTask.SqlServer/Functions.Worker.Extensions.DurableTask.SqlServer.csproj
+++ b/src/Functions.Worker.Extensions.DurableTask.SqlServer/Functions.Worker.Extensions.DurableTask.SqlServer.csproj
@@ -1,0 +1,22 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <!-- Import common properties for all shipping packages-->
+  <Import Project="../common.props" />
+  
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+  </PropertyGroup>
+  
+  <!-- NuGet package settings -->
+  <PropertyGroup>
+    <PackageId>Microsoft.Azure.Functions.Worker.Extensions.DurableTask.SqlServer</PackageId>
+    <Title>Azure Durable Functions SQL Provider</Title>
+    <Description>Microsoft SQL provider for Azure Durable Functions.</Description>
+    <PackageTags>Microsoft;Azure;Functions;Durable;Task;Orchestration;Workflow;Activity;Reliable;SQL</PackageTags>
+  </PropertyGroup>
+
+  <ItemGroup>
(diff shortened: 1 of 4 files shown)


src/Functions.Worker.Extensions.DurableTask.SqlServer/Functions.Worker.Extensions.DurableTask.SqlServer.csproj
```diff
@@ -0,0 +1,22 @@
...
+  <!-- NuGet package settings -->
+  <PropertyGroup>
+    <PackageId>Microsoft.Azure.Functions.Worker.Extensions.DurableTask.SqlServer</PackageId>
+    <Title>Azure Durable Functions SQL Provider</Title>
+    <Description>Microsoft SQL provider for Azure Durable Functions.</Description>
+    <PackageTags>Microsoft;Azure;Functions;Durable;Task;Orchestration;Workflow;Activity;Reliable;SQL</PackageTags>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.Abstractions" Version="1.1.0" />
```
davidmrdavid, line 19: @jviau: can we also add a comment here to update the AssemblyInfo file when this gets changed?

davidmrdavid (approved):
LGTM but left a suggestion

src/Functions.Worker.Extensions.DurableTask.SqlServer/Functions.Worker.Extensions.DurableTask.SqlServer.csproj
jviau: This is not the version that is to be kept in sync. It is the version of Microsoft.Azure.DurableTask.SqlServer.AzureFunctions. Ultimately, it would be nice to automate this - but that can be left for a separate PR as we would need to extract some common versioning targets between the two projects.

davidmrdavid: Ah, my mistake. I misread this (twice, one per PR). Thanks, this makes sense.

cgillum (approved):

cgillum:
We need to also update the docs/quickstart.md file (which currently says .NET Isolated isn't supported) but I can take care of that separately if you prefer.

PR #143, opened 2023-01-19
Fix Az Func dotnet-isolated extension

jviau:
Two fixes:
1. Make sure assembly name and package name match (for consistency)
2. Add `enableImplicitRegistration = true` to the extension attribute.
    - This is because our package has 0 binding attributes in it, the build task skips it unless this is set to `true`.


cgillum (approved):
I verified that this is working!

PR #154, opened 2023-02-23
Add FabricBot config for automatic GitHub triaging

davidmrdavid:
As in title. This will allow FabricBot to manage special labels like "needs: triage" and "needs: author response".

### DIFF
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -1,0 +1,536 @@
+{
+    "version": "1.0",
+    "tasks": [
+        {
+            "taskType": "trigger",
+            "capabilityId": "IssueResponder",
+            "subCapability": "IssuesOnlyResponder",
+            "version": "1.0",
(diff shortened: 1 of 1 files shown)


cgillum:
Wow, that's some very verbose JSON. Is the expectation that we will always have to update this by hand if we want to make changes or is there some UX to help us with this?

davidmrdavid:
@cgillum: this is actually the same config as the one we host in the DF Extension: https://github.com/Azure/azure-functions-durable-extension/blob/dev/.github/fabricbot.json

My understanding is that there's a UX _on the way_ to help manage the config. I'm actually talking with the FabricBot team to try and get onboarded to that UX. That will take time though. For now, I understand the easiest path to onboard these " new repos" is to utilize the same `.json` config as the one we have in the extension.

cgillum (approved):

PR #163, opened 2023-04-08
Support configurable backoff intervals

tompostler:
I was analyzing the cost of an app service with overall low volume (1-2 orchestrations per week), and right now the cost of the app service telemetry outweighs the cost of the basic app service.

Since the orchestrations I run are very tolerable of delays, I wanted to increase the maximum backoff values here to 5 minutes or more for a ~99% reduction in unnecessary telemetry.

As the orchestrations I run are just a hobby project, I'm not as familiar with the potential consequences of changing these values, especially with how they relate to each other. But I found the todo in the repo and figured I would start with the most basic implementation of passing it through on the settings object.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## v1.1.2
 
+### New
+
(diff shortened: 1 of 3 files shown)


cgillum (approved):
Thank you for this contribution! I'd be happy to accept it and include it as part of our next release.

Issue #162, opened 2023-04-05
Explicit taskHub parameter for stored procedures


michaelplavnik:
Please consider changing approach from using implicit dt.CurrentTaskHub() to explicit parameter in all procedures.

Rational. Flexibility in support database sharing/isolation approaches and simplicity.

For example, from documentation: "This is often valuable when your organization has many small apps but prefers to manage only a single backend database. When multitenancy is enabled, different apps connect to a shared database using different database login credentials and each app will only have access to its own data." Explicit assumption of different login is not flexible, and might be not reflecting reality for majority of financial organizations that actually want to share the database schema.

cgillum:
@michaelplavnik can you share more about the setup that you want to enable? There are several ways to share a database and I want to understand if there is a mechanism to enable your scenario that doesn't require explicit task hub parameters.

I anticipate it will be difficult for us to continue support the higher security model if we add explicit task hub parameters.

michaelplavnik:
The following setup: hub process is running under AD account, SQL server trusted authentication, shared connection pool between task hub and domain specific logic. Multiple physical hosts, each running single hub process. Hosts are partitioned based on task hub name.

In regards to your concern. It seems to revolve around @taskHubMode = 1. This can still be enforced even with explicit task hub name (e.g. instead of deducing task hub name, assert USER_NAME() = taskHubName). Other mode then becomes application level responsibility mode.

cgillum:
I see. I think that strategy might work and that the feature could be generally useful - thank you.

While useful, it may be difficult for our team to prioritize this feature at this time without input from more users. That said, many of the features in this project are community contributions, and we'd be happy to accept a PR for this enhancement as well is anyone wants to contribute it (in which case we can publish it immediately).

vigouredelaruse:
i support this line of discussion because operating durable functions with the sql provider drives one into conflict with unforable combinations of at minimum

-    1:1 map of sql-login-name<=> task hub separation. in a dev environment with frequent code changes this is a confounding factor for retiring in-flight orchestrations that encourages bloat of 'abandoned' schemas and databases with inflight orchestrations 
-    tight coupling of schema names and kubernetes func deploy configuration of scalar metric as per the config below. this is a cross-cutting concern that depends on the keda autoscaler configuration

  ``` 
    - metadata:
        connectionStringFromEnv: FunctionStorage_SQLSERVER
        query: SELECT dt.GetScaleRecommendation(10, 1)
        targetValue: '1'
[...]

PR #167, opened 2023-05-02
Explicit taskHub parameter for stored procedures

michaelplavnik:
Resolves issue #162.
Add TaskHubName parameter to all relevant SPs. 
Add parameter to initialize TaskHubMode to the settings class. 
NOTE: Breaking changes in vInstance and vHistory views usage in case of TaskHubMode=0.

### DIFF
--- a/src/DurableTask.SqlServer/SqlOrchestrationService.cs
+++ b/src/DurableTask.SqlServer/SqlOrchestrationService.cs
@@ -37,6 +37,7 @@ namespace DurableTask.SqlServer
         readonly SqlDbManager dbManager;
         readonly string lockedByValue;
         readonly string userId;
+        string? taskHubNameValue;
 
         public SqlOrchestrationService(SqlOrchestrationServiceSettings? settings)
         {
@@ -86,13 +87,18 @@ namespace DurableTask.SqlServer
             return connection;
         }
 
-        SqlCommand GetSprocCommand(SqlConnection connection, string sprocName)
+        SqlCommand GetTaskHubSprocCommand(SqlConnection connection, string sprocName)
         {
-            SqlCommand command = connection.CreateCommand();
-            command.CommandType = CommandType.StoredProcedure;
-            command.CommandText = sprocName;
+            SqlCommand command = connection.GetSprocCommand(string.Concat(this.settings.SchemaName, ".", sprocName));
+            this.AddTaskHubNameParameter(command);
             return command;
         }
+
+        SqlCommand GetFuncCommand(SqlConnection connection, string funcName, SqlDbType retType, int retSize = 0)
+            => connection.GetFuncCommand(string.Concat(this.settings.SchemaName, ".", funcName), retType, retSize);
+
+        void AddTaskHubNameParameter(SqlCommand command) 
+            => command.AddTaskHubNameParameter(this.taskHubNameValue);
 
         public override Task CreateAsync(bool recreateInstanceStore)
             => this.dbManager.CreateOrUpgradeSchemaAsync(recreateInstanceStore);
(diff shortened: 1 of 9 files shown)


michaelplavnik:
@microsoft-github-policy-service agree

cgillum:
@michaelplavnik thanks for submitting this PR. I enabled the CI and it looks like some tests are failing because of inserting `NULL` into the `TaskHub` column.

> Microsoft.Data.SqlClient.SqlException : Cannot insert the value NULL into column 'TaskHub', table 'DurableDB.dt2.Instances'; column does not allow nulls. INSERT fails.

Can you take a look?

michaelplavnik:
@cgillum Thanks for being patient :-) I have update this PR. Could you please run tests again, in my environment I cannot run performance tests.

cgillum (commented):
Tests appear to be passing now. :)

I have a couple concerns about this PR from a breaking-change perspective:

* Adding `@TaskHubName` to all stored procedures - isn't this also a breaking change? Sprocs that don't start with `_` are considered public API and we need to maintain backwards compatibility.
- The modified `CurrentTaskHub` sproc seems to no longer consider `APP_NAME()`. Isn't this also technically a breaking change?

Is there any way to make these changes non-breaking? I'm less concerned about the changes to `vInstances` and `vHistory`.

michaelplavnik:
@cgillum You are correct of cause that changes are breaking (and in case of vInstances, vHistory also hidden). Now, I know that you prefer to keep it backward compatible (vs. explicit breaking change). Let's see if I can make it backward compatible in all cases.

michaelplavnik:
@cgillum Finally got some time to remove braking changes. However, functions now have private implementation and non breaking public wrapper. Please advise if you want public wrapper with explicit parameter as well.

bhugot:
little bump on this @cgillum

src/DurableTask.SqlServer/SqlOrchestrationServiceSettings.cs
```diff
@@ -102,6 +125,14 @@ public SqlOrchestrationServiceSettings(string connectionString, string? taskHubN
         [JsonProperty("createDatabaseIfNotExists")]
         public bool CreateDatabaseIfNotExists { get; set; }
 
+        /// <summary>
+        /// Gets or sets a flag indicating wether task hub is created for multi-tennant naming
+        /// (derived from sql server user name) or for application specified naming.
+        /// </summary>
+        /// <seealso cref="SqlOrchestrationService.CreateAsync(bool)"/>
+        [JsonProperty("createMultiTennantTaskHub ")]
+        public bool CreateMultiTennantTaskHub { get; set; } = true;
```
cgillum, line 134: According to Wikipedia, "[multitenant](https://en.wikipedia.org/wiki/Multitenancy)" is a single word with just one "n". This is also the spelling we use in our documentation. Let's update all the new code in this PR to use that spelling.

Example here:
```suggestion
        public bool CreateMultitenantTaskHub { get; set; } = true;
```

src/DurableTask.SqlServer/SqlOrchestrationServiceSettings.cs
```diff
@@ -14,34 +14,49 @@ namespace DurableTask.SqlServer
...
+
         /// <summary>
         /// Initializes a new instance of the <see cref="SqlOrchestrationServiceSettings"/> class.
         /// </summary>
         /// <param name="connectionString">The connection string for connecting to the database.</param>
         /// <param name="taskHubName">Optional. The name of the task hub. If not specified, a default name will be used.</param>
         /// <param name="schemaName">Optional. The name of the schema. If not specified, the default 'dt' value will be used.</param>
-        public SqlOrchestrationServiceSettings(string connectionString, string? taskHubName = null, string? schemaName = null)
+        /// <param name="connectionStringWithTaskHubName">Optional. When true task hub name is added to connection string.
+        ///  Default value is <c>true</c> to match how connection pool behaved before introdcution of this parameter.</param>
```
cgillum, line 34:
```suggestion
        ///  Default value is <c>true</c> to match how connection pool behaved before introduction of this parameter.</param>
```

src/DurableTask.SqlServer/SqlOrchestrationService.cs
```diff
@@ -19,6 +19,7 @@ namespace DurableTask.SqlServer
     using DurableTask.Core.Query;
     using DurableTask.SqlServer.SqlTypes;
     using DurableTask.SqlServer.Utils;
+    using Dynamitey.DynamicObjects;
```
cgillum, line 22: Is this being used or can it be removed?

cgillum (commented):
Sorry for the delay on this. It was a busy summer. Just a few small things.

src/DurableTask.SqlServer/SqlOrchestrationServiceSettings.cs
michaelplavnik: Fixed

src/DurableTask.SqlServer/SqlOrchestrationServiceSettings.cs
michaelplavnik: Fixed

src/DurableTask.SqlServer/SqlOrchestrationService.cs
michaelplavnik: Fixed

michaelplavnik:
Sorry for a long delay. I have merged latest changes for main. In addition, I have added ability to set connection string as environment variable for integration tests.

PR #169, opened 2023-05-04
Add target based scaling support for MSSQL

bachuv:
This PR adds support for target based scaling. It builds upon https://github.com/Azure/azure-functions-durable-extension/pull/2452. This PR is in draft mode since it hasn't been tested E2E yet.

Changes:
- [x] Add `SqlMetricsProvider` and move `GetMetricsAsync()` code from `SqlScaleMonitor` here
- [x] Add `SqlTargetScaler` (implements `ITargetScaler`)
- [x] Override `TryGetTargetScaler` in SqlDurabilityProvider
- [x] Add unit tests
- [ ] E2E testing


src/DurableTask.SqlServer.AzureFunctions/SqlMetricsProvider.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    class SqlMetricsProvider
+    {
+        readonly SqlOrchestrationService service;
+        readonly int? previousWorkerCount;
+
+        public SqlMetricsProvider(SqlOrchestrationService service, int? previousWorkerCount = null)
```
bachuv, line 14: Adding a note that having `int? previousWorkerCount = null` isn't ideal and will change to get the current worker count.

src/DurableTask.SqlServer.AzureFunctions/SqlMetricsProvider.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        public async Task<SqlScaleMetric> GetMetricsAsync()
+        {
+            // GetRecommendedReplicaCountAsync will write a trace if the recommendation results
+            // in a worker count that is different from the worker count we pass in as an argument.
+            int recommendedReplicaCount = await this.service.GetRecommendedReplicaCountAsync(
+                this.previousWorkerCount,
+                CancellationToken.None);
+
+            return new SqlScaleMetric { RecommendedReplicaCount = recommendedReplicaCount };
```
bachuv, line 28: I moved this code from `GetMetricsAsync()` in `SqlScaleMonitor()`.

Once I moved it here, `previousWorkerCount` is only set at initialization.

I looked at where `previousWorkerCount` is assigned [here ](https://github.com/microsoft/durabletask-mssql/blob/f4527f602bc53376496618f616977474903fc122/src/DurableTask.SqlServer.AzureFunctions/SqlScaleMonitor.cs#L62) and looks like the value comes from `ScaleStatusContext.WorkerCount` [here ](https://github.com/microsoft/durabletask-mssql/blob/f4527f602bc53376496618f616977474903fc122/src/DurableTask.SqlServer.AzureFunctions/SqlScaleMonitor.cs#L48) or [here](https://github.com/microsoft/durabletask-mssql/blob/f4527f602bc53376496618f616977474903fc122/src/DurableTask.SqlServer.AzureFunctions/SqlScaleMonitor.cs#L52). @cgillum, is there any other way to get the current worker count? I'm not sure I understand how the `GetScaleStatus` methods are invoked. Thanks!

src/DurableTask.SqlServer.AzureFunctions/SqlTargetScaler.cs
```diff
@@ -0,0 +1,32 @@
...
+        readonly TargetScalerResult scaleResult;
+
+        public SqlTargetScaler(string functionId, SqlMetricsProvider sqlMetricsProvider)
+        {
+            this.sqlMetricsProvider = sqlMetricsProvider;
+            this.scaleResult = new TargetScalerResult();
+            this.TargetScalerDescriptor = new TargetScalerDescriptor(functionId);
+        }
+
+        public TargetScalerDescriptor TargetScalerDescriptor { get; private set; }
```
cgillum, line 22: nit: remove the `private set` to make this a read-only property. You'll still be able to assign it from the constructor.

src/DurableTask.SqlServer.AzureFunctions/SqlMetricsProvider.cs
cgillum: I'm not 100% sure, but I assume `GetScaleStatus` methods are invoked by the Functions host. Presumably the Functions host already knows about how many workers there are.

src/DurableTask.SqlServer.AzureFunctions/SqlMetricsProvider.cs
bachuv: I realized that `previousWorkerCount` is only used to emit a trace [here](https://github.com/microsoft/durabletask-mssql/blob/1b5ccddd485a9b1da67d54076f77288c1725b5f2/src/DurableTask.SqlServer/SqlOrchestrationService.cs#L888C1-L891) so it seems fine to leave it as a nullable argument.

```
if (currentReplicaCount != null && currentReplicaCount != recommendedReplicaCount)
{
    this.traceHelper.ReplicaCountChangeRecommended(currentReplicaCount.Value, recommendedReplicaCount);
}
```

src/DurableTask.SqlServer.AzureFunctions/SqlDurabilityProvider.cs
```diff
@@ -197,8 +197,24 @@ public override async Task RewindAsync(string instanceId, string reason)
             string storageConnectionString,
             out IScaleMonitor scaleMonitor)
         {
-            scaleMonitor = this.scaleMonitor ??= new SqlScaleMonitor(this.service, hubName);
+            SqlMetricsProvider sqlMetricsProvider = new SqlMetricsProvider(this.service);
+            scaleMonitor = this.scaleMonitor ??= new SqlScaleMonitor(this.service, hubName, sqlMetricsProvider);
             return true;
         }
+
+#if NETCOREAPP
```
cgillum, line 205: Why do we need the `#if NETCOREAPP`? Is there any problem if this code exists in both the `netstandard2.0` and `netcoreapp3.1` targets?

src/DurableTask.SqlServer.AzureFunctions/SqlTargetScaler.cs
```diff
@@ -0,0 +1,32 @@
...
+namespace DurableTask.SqlServer.AzureFunctions
+{
+    using System.Threading.Tasks;
+    using Microsoft.Azure.WebJobs.Host.Scale;
+
+#if !NETSTANDARD
+    public class SqlTargetScaler : ITargetScaler
+    {
+        readonly SqlMetricsProvider sqlMetricsProvider;
+        readonly TargetScalerResult scaleResult;
```
cgillum, line 13: Why do we cache the `TargetScalerResult` object instead of creating a new one for each call?

src/DurableTask.SqlServer.AzureFunctions/SqlDurabilityProvider.cs
cgillum: I figured this out. This is necessary because our dependency chain doesn't give us access to `ITargetScaler` when targeting `netstandard2.0`.

src/DurableTask.SqlServer.AzureFunctions/SqlTargetScaler.cs
cgillum: I went ahead and removed this caching. I'm worried about potential race condition side effects that might happen with shared state.

cgillum:
Quick note: I've successfully tested these changes in the Azure Functions SCv3 integration test code.

src/DurableTask.SqlServer.AzureFunctions/DurableTask.SqlServer.AzureFunctions.csproj
```diff
@@ -4,7 +4,11 @@
   <Import Project="../common.props" />
   
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net6.0'">
+    <DefineConstants>$(DefineConstants);FUNCTIONS_V4</DefineConstants>
```
davidmrdavid, line 11: I think bundles still targets TFM netcoreapp3.1.
But I know the host targets net6.0

I hope that still means that at runtime the FUNCTIONS_v4 variable will be defined. Not super sure about the resolution rules here. But wanted to call this out.

src/DurableTask.SqlServer.AzureFunctions/SqlMetricsProvider.cs
```diff
@@ -0,0 +1,29 @@
...
+namespace DurableTask.SqlServer.AzureFunctions
+{
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    public class SqlMetricsProvider
+    {
+        readonly SqlOrchestrationService service;
+
+        public SqlMetricsProvider(SqlOrchestrationService service, int? previousWorkerCount = null)
```
davidmrdavid, line 13: is `previousWorkerCount` unused? Do we need it in the constructor at all?

src/DurableTask.SqlServer.AzureFunctions/SqlScaleMonitor.cs
```diff
@@ -19,14 +19,22 @@ class SqlScaleMonitor : IScaleMonitor<SqlScaleMetric>
...
         int? previousWorkerCount = -1;
 
-        public SqlScaleMonitor(SqlOrchestrationService service, string taskHubName)
+        public SqlScaleMonitor(string taskHubName, SqlMetricsProvider sqlMetricsProvider)
         {
-            this.service = service ?? throw new ArgumentNullException(nameof(service));
-            this.Descriptor = new ScaleMonitorDescriptor($"DurableTask-SqlServer:{taskHubName ?? "default"}");
+            // Scalers in Durable Functions are shared for all functions in the same task hub.
+            // So instead of using a function ID, we use the task hub name as the basis for the descriptor ID.
+            string id = $"DurableTask-SqlServer:{taskHubName ?? "default"}";
```
davidmrdavid, line 30: huh, I think we've been getting this wrong across the board. This approach - one monitor per taskhub instead of a monitor per functionID, seems like what we should have done all along. If your tests show this works, that's great. I just hope there's no hidden dependency on the funcitonID being reflected on the monitor's own ID. 
Taking note.

test/DurableTask.SqlServer.AzureFunctions.Tests/TargetBasedScalingTests.cs
```diff
@@ -0,0 +1,59 @@
...
+        readonly Mock<IOrchestrationService> orchestrationServiceMock;
+
+        public TargetBasedScalingTests()
+        {
+            this.orchestrationServiceMock = new Mock<IOrchestrationService>(MockBehavior.Strict);
+
+            this.metricsProviderMock = new Mock<SqlMetricsProvider>(
+                MockBehavior.Strict,
+                null,
+                null);
```
davidmrdavid, line 27: non-blocking heads up: but I'm a fan of using named parameters whenever `null` is used in a method call so that a reader can know what parameters are being set. Otherwise, it's a bit difficult to examine.

But for this review, don't worry about it.

test/DurableTask.SqlServer.AzureFunctions.Tests/TargetBasedScalingTests.cs
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        [Theory]
+        [InlineData(0)]
+        [InlineData(10)]
+        [InlineData(20)]
+        public async void TargetBasedScalingTest(int expectedTargetWorkerCount)
```
davidmrdavid, line 34: is it theoretically possible for the expectedWorkerCount to return a negative number? I guess that the `int` type doesn't have any prevention against it. In the spirit of defensive programming, should we have a guard against that edge case?

davidmrdavid (commented):
Some questions, but overall looks great!

src/DurableTask.SqlServer.AzureFunctions/DurableTask.SqlServer.AzureFunctions.csproj
cgillum: V4 bundles were changed to target net6.0 a few months ago, so we should be good to also target net6.0: https://github.com/Azure/azure-functions-extension-bundles/pull/407.

src/DurableTask.SqlServer.AzureFunctions/SqlScaleMonitor.cs
cgillum: The MSSQL backend has been doing this always, and it's something I discovered when looking closely at the Azure Storage scale logs in the past (though for some reasons I never got around to suggesting we fix it). Basically, the existing scaler logic is creating redundant scalers that are all doing the exact same thing with the exact same data - which still works but is wasteful and can be costly to the customer, who gets billed for all these storage transactions.

test/DurableTask.SqlServer.AzureFunctions.Tests/TargetBasedScalingTests.cs
cgillum: Shouldn't be possible, based on the code [here](https://github.com/microsoft/durabletask-mssql/blob/6003e41398779f7796aead207e399d3031c6e623/src/DurableTask.SqlServer/Scripts/logic.sql#L147-L167). But I'm not opposed to adding some defensive code.

test/DurableTask.SqlServer.AzureFunctions.Tests/TargetBasedScalingTests.cs
cgillum: Same! Fixed'ish. One awkward thing in this case is that this is a `params` parameter, which means I can't use named parameters in the way that you would with a normal method call.

src/DurableTask.SqlServer.AzureFunctions/SqlMetricsProvider.cs
cgillum: Removed.

davidmrdavid (approved):

Issue #161, opened 2023-03-28
Exception is not propagated to parent orchestration as expected


moldovangeorge:
Irrespective of the [ErrorPropagationMode](https://github.com/Azure/durabletask/blob/c2c51c2ab9411eb0e440b4c508985e3d53c2e14e/src/DurableTask.Core/ErrorPropagationMode.cs ) value that is set on the worker when running based on the SqlServerProvider, the Exception is not propagated up the stack as expected. 
My expectation was that the Exception that is caught in the parent orchestration would contain the exception that was thrown in the Activity either in the innerException field, or in the FailureDetails field. 
![image](https://user-images.githubusercontent.com/73134867/228214917-702dc566-c477-49d4-bef6-b95bbb7ec50e.png)
I ran a demo app using both UseFailureDetails and UseFailureDetails setup and in both cases the exception was not propagated correctly, and it only contained limited information in the message of the exception ( either the message of the inner exception, or the type of the inner exception, depending if the inner exception had a message or not). 

Also, when using the UseFailureDetails setup, I expected to receive an OrchestrationException in parent orchestration but a TaskFailedException was raised instead.

wros:
I have this same problem using the function, in isolated mode. The Exception is not propagated to the orchestration, and I cannot use the custom retry in the activity.

I'm using the Azure Version.

When I use the SQL Provider

![image](https://github.com/microsoft/durabletask-mssql/assets/26489708/0d4f48e0-98fc-4bc9-8095-9bf43104b830)

Here when I don't use the SQL Provider:

![image](https://github.com/microsoft/durabletask-mssql/assets/26489708/7798d59f-b189-4341-bef8-ef017ab0c571)

Obs: I'm using the VS 2022 default template for Azure Functions.

cgillum:
Thanks @wros for the additional context. This helps narrow things down. I haven't had a chance to look at this yet but hoping to get to it soon'ish.

cgillum:
I spent some time debugging this and I think the problem is that the database schema needs to be updated to support correctly persisting detailed exception information. From what I can tell, we're only saving the exception message, which isn't something we can correctly deserialize back into code. This is definitely a missing feature.

PR #188, opened 2023-09-25
Fix error handling propagation

cgillum:
Resolves https://github.com/microsoft/durabletask-mssql/issues/161

For whatever reason, we forgot to fully implement exception propagation in this MSSQL backend for the Durable Task Framework. Fundamentally, the reason this is easy to miss is because the MSSQL provided doesn't serialize history events, but rather manually reads and writes them into structured SQL tables.

In this PR, we implement handling for both `ErrorPropagationMode.SerializeExceptions` (which we call "legacy") and `ErrorPropagationMode.UseFailureDetails` (which I'm treating as the new default).

To summarize the changes in this PR, we:
* Make sure to read/write failure details to and from the database as JSON
* Make sure we surface error details in the right way in the public API
* Add new tests or improve existing tests to verify that we flow exception details in expected ways

### DIFF
--- a/src/DurableTask.SqlServer/DTUtils.cs
+++ b/src/DurableTask.SqlServer/DTUtils.cs
@@ -5,9 +5,11 @@ namespace DurableTask.SqlServer
 {
     using System;
     using System.Diagnostics;
+    using System.IO;
     using System.Text;
     using DurableTask.Core;
     using DurableTask.Core.History;
+    using Newtonsoft.Json;
     using SemVersion;
 
     static class DTUtils
@@ -17,6 +19,13 @@ namespace DurableTask.SqlServer
 
         // DurableTask.Core has a public static variable that contains the app name
         public static readonly string AppName = DurableTask.Core.Common.Utils.AppName;
+
+        static readonly JsonSerializer DefaultJsonSerializer = JsonSerializer.Create(
+            new JsonSerializerSettings
+            {
+                TypeNameHandling = TypeNameHandling.None,
+                NullValueHandling = NullValueHandling.Ignore,
+            });
 
         static SemanticVersion GetExtensionVersion()
         {
@@ -98,18 +107,14 @@ namespace DurableTask.SqlServer
             }
         }
 
-        public static int GetPayloadSizeInBytes(TaskMessage message)
-        {
-            if (TryGetPayloadText(message.Event, out string? payloadText) && payloadText != null)
-            {
-                return Encoding.UTF8.GetByteCount(payloadText);
-            }
-
-            return -1;
-        }
-
         public static bool TryGetPayloadText(HistoryEvent e, out string? payloadText)
         {
+            if (TryGetFailureDetails(e, out FailureDetails? failureDetails) && failureDetails != null)
+            {
+                payloadText = SerializeToJson(failureDetails);
+                return true;
+            }
+
             payloadText = e.EventType switch
             {
                 EventType.ContinueAsNew => ((ContinueAsNewEvent)e).Result,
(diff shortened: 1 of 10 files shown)


src/DurableTask.SqlServer/DTUtils.cs
```diff
@@ -180,5 +200,52 @@ public static bool HasPayload(HistoryEvent e)
...
             };
         }
+
+        public static bool TryDeserializeFailureDetails(string jsonText, out FailureDetails? failureDetails)
+        {
+            try
+            {
+                failureDetails = DeserializeFromJson<FailureDetails>(jsonText);
+
+                // Depending on how the TaskHubWorker is configured, the text might be a FailureDetails JSON object
```
bachuv, line 210: Just curious, can you share a link to the code where the `ErrorPropagationMode` is set to the default of `ErrorPropagationMode.UseFailureDetails`?

bachuv (approved):
Looks good to me! Just left a small question.

src/DurableTask.SqlServer/DTUtils.cs
cgillum: In this project, the default is set up in TestService.cs, line 68. The default mode can be set in the constructor parameter. For Orchestrations.cs, we don't specify anything, which causes the default value to be `UseFailureDetails`. Only in LegacyErrorPropagation.cs do we use the `SerializeExceptions` mode.

FWIW, the property is defined here: [`TaskHubWorker.ErrorPropagationMode`](https://github.com/Azure/durabletask/blob/6dc7134ceee4cec9073b86d5bb70da0cd4ca6731/src/DurableTask.Core/TaskHubWorker.cs#L130-L145). We don't set the default value explicitly, but enum properties default to zero, which in this case is [`SerializeExceptions`](https://github.com/Azure/durabletask/blob/6dc7134ceee4cec9073b86d5bb70da0cd4ca6731/src/DurableTask.Core/ErrorPropagationMode.cs#L28). This is the default because that was always the original behavior of the Durable Task Framework and changing the default behavior could break in-flight orchestrations for existing apps. It was much later that we added support for `UseFailureDetails`.

cgillum: Not sure if this is what you were asking, but we also set it in the .NET Isolated SDK [here](https://github.com/microsoft/durabletask-dotnet/blob/e2f329258d1e02a762911bd89274a9a584b6ec04/src/Worker/Grpc/GrpcDurableTaskWorker.Processor.cs#L213-L217).

Java assumes this behavior as well, but doesn't require any explicit configuration since Java doesn't rely on the .NET implementation of `TaskHubWorker`. Basically, this is the _only_ option for Java.

bachuv: Thanks for the explanation! I understand it a lot better now.

aaditya8:
Hi, We are waiting for this fix from so long and upgraded sqlserver libraries and core libraries and started testing with both the modes "ErrorPropagationMode.SerializeExceptions", "ErrorPropagationMode.UseFailureDetails". Please find my observations help me with my use case. 

**Use Case :** 

We have one parent orchestration , one sub orchestration under parent orchestration , one task(Activity) under sub orchestration. 
I am expecting to propagate the exception occurred in sub orchestration task to both Sub orchestration and Parent orchestration. Below are my findings.

**Scenarios :**

1. With SerializeExceptions : Inner Exception is always null in Sub Orchestration and Parent Orchestration.
2. With UseFailureDetails : I am able to see the Failure details(Including stack trace) of the Sub Orchestration Activity in Sub Orchestration. Same way i am able to see the failure details of Sub Orchestration in Parent Orchestration but not able to see the Failure details of Sub orchestration activity. Its coming as null. 

Could you please help me with this.

cgillum:
@aaditya8 please open a new issue for this and include details such as the kind of exceptions that are being raised. In that new issue, please also share code that shows how you're setting up your `TaskHubWorker`.

PR #190, opened 2023-09-28
Distributed tracing support

cgillum:
This PR adds compatibility for the distributed tracing work that was recently done in DurableTask.Core.

Summary of changes:

- Added TraceContext DB schema objects as part of new 1.2.0 schema version
- Updated DB management tests to verify the correct version of the schema is listed in the DB
- Added new OTel distributed tracing integration test
- Updated xunit test dependencies (which included some breaking test changes)
- Updated nuget package versions to 1.2.0


src/DurableTask.SqlServer/Scripts/logic.sql
```diff
@@ -1,6 +1,85 @@
...
+        [TaskID] int NULL,
+        [Timestamp] datetime2 NULL,
+        [IsPlayed] bit NULL,
+        [VisibleTime] datetime2 NULL,
+        [Reason] varchar(max) NULL,
+        [PayloadText] varchar(max) NULL,
+        [PayloadID] uniqueidentifier NULL,
+        [ParentInstanceID] varchar(100) NULL,
+        [Version] varchar(100) NULL,
+        [TraceContext] varchar(800) NULL
```
cgillum, line 39: Each of these type definitions previously lived in schema-1.0.0.sql. I realized, however, that they needed to be moved to logic.sql so that we can more easily update them when changes are needed (like adding `TraceContext` properties). The only changes are to HistoryEvents, OrchestrationEvents, and TaskEvents, where we add the `TraceContext` property to the end of the list of properties.

src/DurableTask.SqlServer/Scripts/schema-1.2.0.sql
```diff
@@ -0,0 +1,33 @@
...
+
+-- PERSISTENT SCHEMA OBJECTS (tables, indexes, types, etc.)
+--
+-- The contents of this file must never be changed after
+-- being published. Any schema changes must be done in
+-- new schema-{major}.{minor}.{patch}.sql scripts.
+
+-- Add a new TraceContext column to the Instances table
+IF NOT EXISTS (SELECT 1 FROM sys.columns WHERE object_id = OBJECT_ID('__SchemaNamePlaceholder__.Instances') AND name = 'TraceContext')
+    ALTER TABLE __SchemaNamePlaceholder__.Instances ADD TraceContext varchar(800) NULL
```
cgillum, line 12: This is the first release where we have a database schema upgrade! Schema upgrades must be done by adding a new `schema-X.Y.Z.sql` file, and each file only contains changes, like adding columns.

test/DurableTask.SqlServer.Tests/Integration/DatabaseManagement.cs
```diff
@@ -319,19 +326,16 @@ public async Task CanCreateIfNotExists(bool isDatabaseMissing)
...
             using TestDatabase testDb = this.CreateTestDb(!isDatabaseMissing);
             IOrchestrationService service = this.CreateServiceWithTestDb(testDb);
 
             // Simulate 4 workers starting up concurrently and trying to initialize
             // the same database schema. It should just work with predictable output.
-            Parallel.For(0, 4, i =>
-            {
-                service.CreateIfNotExistsAsync().GetAwaiter().GetResult();
-            });
+            await Enumerable.Range(0, 4).ParallelForEachAsync(4, i => service.CreateIfNotExistsAsync());
```
cgillum, line 336: The xunit change added some new analyzers that raise compiler errors if we do things like block threads with calls like `Task.GetAwaiter().GetResult()`. That's why I had to change this line of code. It should be mostly equivalent to the previous `Parallel.For` call.

test/DurableTask.SqlServer.Tests/Integration/Orchestrations.cs
```diff
@@ -631,14 +637,136 @@ public async Task RecreateRunningInstance()
...
             string oldExecutionId = instance.ExecutionId;
             await instance.RestartAsync(input, new[] { OrchestrationStatus.Canceled });
             Assert.NotEqual(oldExecutionId, instance.ExecutionId);
         }
+
+        [Fact]
+        public async Task TraceContextFlowCorrectly()
+        {
+            string traceSourceName = "MyTraceSource";
+            string orchestrationName = "";
```
cgillum, line 654: Whoops - just now noticing that I'm using the empty string as the orchestrator name (which explains some confusing logs I was looking at yesterday). Will fix this.

test/DurableTask.SqlServer.Tests/Utils/TestService.cs
```diff
@@ -205,6 +206,14 @@ public void RegisterInlineActivity(string name, string version, TaskActivity act
             this.worker.AddTaskActivities(new TestObjectCreator<TaskActivity>(name, version, activity));
         }
 
+        public void RegisterInlineOrchestration<TOutput, TInput>(
+            string orchestrationName,
+            Func<OrchestrationContext, TInput, Task<TOutput>> implementation,
+            Action<OrchestrationContext, string, string> onEvent = null)
```
cgillum, line 212: This is a new convenience overload that doesn't require the caller to pass the `version` parameter (which we rarely use).

cgillum (commented):
Added some PR comments to give more context on the changes.

cgillum:
The `MultiInstanceQueries` test is failing consistently. I'll take a look at why it's failing and push an update as appropriate.

bachuv (commented):
Thanks for opening this PR with the distributed tracing changes! The comments you added in the PR were useful to refer to when reviewing it!

src/DurableTask.SqlServer/Utils/PartitionedQueue.cs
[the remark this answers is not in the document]
cgillum: It was never actually being used, and I had forgotten about it. After upgrading to the new xunit, tests related to this file started having errors, and I remembered that this file needed to be removed.

src/DurableTask.SqlServer/Scripts/schema-1.2.0.sql
cgillum: Yes, it's very similar to the Antares database upgrade model.

test/DurableTask.SqlServer.Tests/Integration/DatabaseManagement.cs
```diff
@@ -490,6 +495,16 @@ static void ValidateDatabaseSchema(TestDatabase database, string schemaName = "d
...
+
+            // Verify that the schema version in the database matches the expected version
+            // Note that we'll need to change the expected version here whenever we introduce new schema.
+            SemanticVersion currentSchemaVersion = await SharedTestHelpers.GetCurrentSchemaVersionAsync(
+                this.output,
+                database.ConnectionString,
+                schemaName);
+            Assert.Equal(1, currentSchemaVersion.Major);
+            Assert.Equal(2, currentSchemaVersion.Minor);
+            Assert.Equal(0, currentSchemaVersion.Patch);
```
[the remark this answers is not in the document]
cgillum: This isn't ideal, but I wasn't able to think of a better process. The version is supposed to match the nuget package version, but we currently only update the schema version when we actually make changes, so this number is going to always be equal or less than the nuget package version.

My expectation, however, is that the next time we update the schema, the schema version will be updated automatically, and this test will fail. This failure will alert us to update this value, and the person updating it will be able to make sure that the database schema version matches what we intuitively expect. People will be forced to update this in order to get the test passing again.

src/DurableTask.SqlServer/Scripts/logic.sql
```diff
@@ -1,6 +1,85 @@
 ﻿-- Copyright (c) Microsoft Corporation.
 -- Licensed under the MIT License.
 
+-- Create custom types. This must be done before creating stored procedures.
```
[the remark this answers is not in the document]
cgillum: Whenever the app starts up, it checks the version of the schema in the database, which is stored in a `dt.Versions` table. If it doesn't find anything (i.e. a new database), it runs these scripts. The order is `schema-x.y.z.sql` (in logical order), `logic.sql` and finally `permissions.sql`. If it finds a version that's older than the newest version of the `schema-x.y.z.sql` version, then it runs all the higher-versioned `schema-x.y.z.sql` files and then runs `logic.sql`. The latter case is what will happen for existing users that upgrade to the `1.2.0` nuget package.

I guess it would make sense to put this in a README.md file in this directory.

cgillum (commented):
Responding to comments.

src/DurableTask.SqlServer/Scripts/logic.sql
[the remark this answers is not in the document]
bachuv: Thanks for the explanation!

src/DurableTask.SqlServer/Scripts/schema-1.2.0.sql
cgillum: I also added a new README.md file in the latest commit to explain the process in more detail. That will probably be easier than studying old PRs (though that will still be an option).

PR #194, opened 2023-10-03
Fix minor issue with DateTimeKind

cgillum:
There was a minor issue I found while doing some prototyping where some serializers (like protobuf) don't accept DateTime values which aren't explicitly set to UTC. This is possible in `OrchestrationState` when we use `default` for a `DateTime` that doesn't have a valid value.

The fix is to ensure that all `DateTime` objects we return, including `default` have a `DateTimeKind` of UTC. In the process, I also simplified some of the helper methods for reading DateTime values from the database.


cgillum:
Fixed as part of https://github.com/microsoft/durabletask-mssql/pull/195.

Issue #181, opened 2023-08-04
SqlUtils doesn't support suspending


Kobudzik:
Hello there,

I'm working on managing orchestration instances for my organization. Termination operation works fine, but suspending is not supported in SqlUtils.cs (and never was) - hence `System.InvalidOperationException: Don't know how to interpret 'ExecutionSuspended'` is thrown during history event handling. Is there any reason why it's not there, or was it just missed?

cgillum:
Thanks for reporting this! We'll look into adding support for suspended orchestrations.

mattj-MS:
Just wanted to comment this is killing us as well.

cgillum:
Sorry that this fell off the radar. I've submitted a PR with a fix: https://github.com/microsoft/durabletask-mssql/pull/195.

cgillum:
The fix for this has been released: https://github.com/microsoft/durabletask-mssql/releases/tag/v1.2.1

PR #195, opened 2023-11-02
Implement support for suspend/resume

cgillum:
Fixes https://github.com/microsoft/durabletask-mssql/issues/181

The MSSQL backend was missing support for suspend/resume. It also made some bad assumptions about how it would eventually behave when introduced, so some existing logic needed to be removed.

I also bundled a couple additional minor fixes into this PR:

* Fixed inconsistent exception throwing in SqlOrchestrationService.WaitForOrchestrationAsync (realized this was a problem while writing my suspend/resume test)
* Fixed minor issue with default DateTime handling (force it to always be UTC) - this is an issue with certain kinds of serializers, like protobuf, which reject DateTime values that aren't explicitly UTC (found this doing some unrelated exploratory prototyping).

### DIFF
--- a/src/DurableTask.SqlServer/Scripts/logic.sql
+++ b/src/DurableTask.SqlServer/Scripts/logic.sql
@@ -630,7 +630,6 @@ BEGIN
             E.[InstanceID] = I.[InstanceID]
     WHERE
         I.TaskHub = @TaskHub AND
-        I.[RuntimeStatus] NOT IN ('Suspended') AND
 	    (I.[LockExpiration] IS NULL OR I.[LockExpiration] < @now) AND
         (E.[VisibleTime] IS NULL OR E.[VisibleTime] < @now)
 
--- a/src/DurableTask.SqlServer/SqlOrchestrationService.cs
+++ b/src/DurableTask.SqlServer/SqlOrchestrationService.cs
@@ -205,6 +205,7 @@ namespace DurableTask.SqlServer
                             currentStatus = SqlUtils.GetRuntimeStatus(reader);
                             isRunning =
                                 currentStatus == OrchestrationStatus.Running ||
+                                currentStatus == OrchestrationStatus.Suspended ||
                                 currentStatus == OrchestrationStatus.Pending;
                         }
                         else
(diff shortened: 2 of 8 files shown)


src/DurableTask.SqlServer/Scripts/logic.sql
```diff
@@ -630,7 +630,6 @@ BEGIN
             E.[InstanceID] = I.[InstanceID]
     WHERE
         I.TaskHub = @TaskHub AND
-        I.[RuntimeStatus] NOT IN ('Suspended') AND
```
cgillum, line 633: This was added long before we actually implemented suspend/resume in DurableTask.Core, and was based on the assumption that we wouldn't even load suspended orchestrations. In the final design, however, we decided that we needed to continue loading suspended orchestrations, so this line needed to be removed.

src/DurableTask.SqlServer/SqlOrchestrationService.cs
```diff
@@ -570,19 +571,7 @@ public override async Task SendTaskOrchestrationMessageAsync(TaskMessage message
...
-                catch (TaskCanceledException)
-                {
-                    if (timeoutCts.Token.IsCancellationRequested)
-                    {
-                        throw new TimeoutException($"A caller-specified timeout of {timeout} has expired, but instance '{instanceId}' is still in an {state?.OrchestrationStatus.ToString() ?? "unknown"} state.");
-                    }
-
-                    throw;
-                }
+                await Task.Delay(TimeSpan.FromSeconds(1), combinedCts.Token);
```
cgillum, line 574: The above code was deleted to ensure that we only throw `OperationCanceledException` (or `TaskCanceledException`) if this times out. Otherwise, sometimes we'd throw `TimeoutException` and sometimes we'd throw `OperationCanceledException`, depending on which line of code was running when the cancellation token was signaled.

test/DurableTask.SqlServer.Tests/Utils/TestInstance.cs
```diff
@@ -71,9 +71,13 @@ public async Task<OrchestrationState> WaitForStart(TimeSpan timeout = default)
             OrchestrationStatus expectedStatus = OrchestrationStatus.Completed,
             object expectedOutput = null,
             string expectedOutputRegex = null,
-            bool continuedAsNew = false)
+            bool continuedAsNew = false,
+            bool doNotAdjustTimeout = false)
         {
-            AdjustTimeout(ref timeout);
+            if (!doNotAdjustTimeout)
```
cgillum, line 77: Needed to add this, otherwise we can't write test code that waits for timeouts (like I do in my new test) when a debugger is attached.

cgillum (commented):
Adding a few comments with additional context.

bachuv (approved):
LGTM!

Issue #198, opened 2023-11-25
NewEvents stucks


bhugot:
Hello, 

While watching our process I see that some CompletedTask events stay in NewEvents table for really long time (hours)
while other events are removed quite fast (for same type of orchestration).  They are finally processed.

As I already posted we also met 2 "duplications" problems (sometimes due to a restart of an instance).

1.  2 events for same task for same instanceId
2.  1 events for an already completed task (we can see this in appinsight )
    

>  System.InvalidCastException: Unable to cast object of type 'DurableTask.Core.History.TaskCompletedEvent' to type 'DurableTask.Core.History.TaskScheduledEvent'.
   at DurableTask.Core.TaskOrchestrationDispatcher.ReconcileMessagesWithState(TaskOrchestrationWorkItem workItem, String dispatcher, ErrorPropagationMode errorPropagationMode, LogHelper logHelper) in /_/src/DurableTask.Core/TaskOrchestrationDispatcher.cs:line 860
   at DurableTask.Core.TaskOrchestrationDispatcher.OnProcessWorkItemAsync(TaskOrchestrationWorkItem workItem) in /_/src/DurableTask.Core/TaskOrchestrationDispatcher.cs:line 345
   at DurableTask.Core.TaskOrchestrationDispatcher.OnProcessWorkItemAsync(TaskOrchestrationWorkItem workItem)
   at DurableTask.Core.TaskOrchestrationDispatcher.OnProcessWorkItemSessionAsync(TaskOrchestrationWorkItem workItem) in /_/src/DurableTask.Core/TaskOrchestrationDispatcher.cs:line 211
   at DurableTask.Core.WorkItemDispatcher`1.ProcessWorkItemAsync(WorkItemDispatcherContext context, Object workItemObj) in /_/src/DurableTask.Core/WorkItemDispatcher.cs:line 373

cgillum:
Thanks for reporting this. I was able to do a search and confirm that we're seeing some events like this in production. More investigation is needed to determine the root cause.

cgillum:
I found the issue for the exception you are seeing. The two linked PRs should fix it.

bhugot:
Hello @cgillum Are you releasing it today?

cgillum:
@bhugot yes, this is released now: https://github.com/microsoft/durabletask-mssql/releases/tag/v1.2.2.

[...]

PR #201, opened 2023-12-16
Fix SQL roundtrip issue for history event IDs

cgillum:
Fixes https://github.com/microsoft/durabletask-mssql/issues/198.

The other issue that resulted in this InvalidCastException is caused by https://github.com/Azure/durabletask/pull/1018. This PR removes the root cause of the issue, which was that we're not correctly round-tripping event IDs for some history events (one of the downsides of custom event serialization). DurableTask.Core was making certain assumptions about the values of event IDs that weren't valid due to this bug.

### DIFF
--- a/src/DurableTask.SqlServer/SqlUtils.cs
+++ b/src/DurableTask.SqlServer/SqlUtils.cs
@@ -137,7 +137,7 @@ namespace DurableTask.SqlServer
                     historyEvent = new SubOrchestrationInstanceCompletedEvent(eventId: -1, GetTaskId(reader), GetPayloadText(reader));
                     break;
                 case EventType.SubOrchestrationInstanceCreated:
-                    historyEvent = new SubOrchestrationInstanceCreatedEvent(eventId: -1)
+                    historyEvent = new SubOrchestrationInstanceCreatedEvent(eventId)
                     {
                         Input = GetPayloadText(reader),
                         InstanceId = "", // Placeholder - shouldn't technically be needed (adding it requires a SQL schema change)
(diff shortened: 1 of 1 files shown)


src/DurableTask.SqlServer/SqlUtils.cs
```diff
@@ -134,7 +134,7 @@ public static HistoryEvent GetHistoryEvent(this DbDataReader reader, bool isOrch
                     historyEvent = new OrchestratorStartedEvent(eventId);
                     break;
                 case EventType.SubOrchestrationInstanceCompleted:
-                    historyEvent = new SubOrchestrationInstanceCompletedEvent(eventId, GetTaskId(reader), GetPayloadText(reader));
+                    historyEvent = new SubOrchestrationInstanceCompletedEvent(eventId: -1, GetTaskId(reader), GetPayloadText(reader));
```
cgillum, line 137: Specifying -1 for the event ID is how these events are normally created in DurableTask.Core. The fact that were were initializing these events with different values was confusing certain code within DurableTask.Core which was overly sensitive to their specific values.

bachuv: Can you share a link to the code in DurableTask.Core where the event ID confusion happened?

cgillum: Sure - it's the code in [the other PR: TaskOrchestrationDispatcher.ReconcileMessagesWithState](https://github.com/Azure/durabletask/pull/1018/files).

Basically, we were iterating through the orchestration history looking for an event with a particular ID and then casting it to a particular type. Normally this works fine because each event in the history has a unique event ID number (except for the ones with -1). The DT.Core code assumed that EventId values would be unique. However, the bug fixed by this PR was that we were setting multiple (related) events with the same Event ID value, which is why we were casting history events to the wrong type as shown in the error message for that PR.

bachuv (approved):
LGTM!

PR #210, opened 2024-03-27
Updates for v1.2.3 release.

cgillum:
Updating package versions from 1.2.2 to 1.2.3.

Plus a couple other very small changes.

### DIFF
--- a/src/common.props
+++ b/src/common.props
@@ -30,6 +30,7 @@
   <PropertyGroup>
     <Authors>Microsoft</Authors>
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
+    <PackageIcon>logo.png</PackageIcon>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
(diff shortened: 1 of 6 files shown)


src/common.props
```diff
@@ -30,7 +30,6 @@
   <PropertyGroup>
     <Authors>Microsoft</Authors>
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
-    <PackageIcon>logo.png</PackageIcon>
```
davidmrdavid, line 33: no logo? :(

davidmrdavid (approved):
LGTM

src/common.props
cgillum: I unfortunately made this change a long time ago but hadn't yet pushed it, and now I can't remember exactly why I did this. I suspect I did it due to some warning, but I can't remember what the warning was and can't find any documentation explaining why having this would be a problem. I'll revert this change and revisit later.

PR #218, opened 2024-05-07
Issue217 - Rollback transaction in CreateInstance stored proc to prevent deadlocks

microrama:
Issue217 - Rollback transaction before throwing error when an orchestration instance has already been created by another process/thread. This is to prevent deadlocks as explained in the issue.


microrama:
@microsoft-github-policy-service agree [company="{MTB}"]

microrama:
@microsoft-github-policy-service agree
@microsoft-github-policy-service agree company="MTB"

cgillum (approved):

bhugot:
I didn t had the time today but did you check that was the only case?

PR #221, opened 2024-05-23
Fix SqlUtils.WithRetry() to open a new connection if the last attempt closed the connection due to network issues.

microrama:
Pass the DBCommand to the SqlUtils.WithRetry() method and let the method check the DBCommand connection state and open if needed before attempting the next attempt. This way, we avoid throwing an Error back when there is a intermittent network connectivity issue.


cgillum (approved):
LGTM!

PR #223, opened 2024-05-30
Add support for specifying schemaName through environment variables

davidmrdavid:
We've received a request to allow the schemaName to be specified via environment variables. This PR aims to add support for that. Still WIP


cgillum:
The proper way to do this is to use `INameResolver.ResolveWholeString`. Here's an example: https://github.com/microsoft/durabletask-mssql/commit/dd40481881f2cc608f95b2717727ea63d9a73c92.

PR #226, opened 2024-05-31
Clear NuGet security warnings

davidmrdavid:
This is the DTFx-MSSQL version of this PR: https://github.com/Azure/azure-functions-durable-extension/pull/2842

As part of the 1ES migration, we have new supply chain requirements around our repo's nuget.config files.

In particular, we're getting errors due to use MyGet. This PR removes our MyGet usage. It also addresses warnings regarding missing `<clear/>` statements.

### DIFF
--- a/nuget.config
+++ b/nuget.config
@@ -4,6 +4,5 @@
   <packageSources>
     <clear/>
     <add key="nuget.org" value="https://www.nuget.org/api/v2/" />
-    <add key="azure-appservice" value="https://www.myget.org/F/azure-appservice/api/v3/index.json" />
   </packageSources>
 </configuration>
--- a/nuget.config
+++ b/nuget.config
@@ -2,8 +2,8 @@
 
 <configuration>
   <packageSources>
+    <clear/>
     <add key="nuget.org" value="https://www.nuget.org/api/v2/" />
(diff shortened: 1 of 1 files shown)


nuget.config
```diff
@@ -2,8 +2,8 @@
 
 <configuration>
   <packageSources>
+    <clear/>
     <add key="nuget.org" value="https://www.nuget.org/api/v2/" />
-    <add key="azure-appservice-staging" value="https://www.myget.org/F/azure-appservice-staging/api/v3/index.json" />
     <add key="azure-appservice" value="https://www.myget.org/F/azure-appservice/api/v3/index.json" />
```
jviau, line 7: this will need to be removed or replaced with an ADO feed as well.

jviau (changes requested):

nuget.config
davidmrdavid: ah, I missed there were two of these. Fixing promptly

davidmrdavid: @jviau - fixed in next commit

davidmrdavid: actually, I guess that if we only have `nuget.org`, we don't really need a `nuget.config`....but I think we'll eventually want to add the ADO feed so I'll keep it around

jviau: You are expected to always have a `nuget.config`, because the `<clear />` element is important.

jviau (approved):

PR #231, opened 2024-09-03
Update taskhubs.md with correct storageProvider type value

cgillum:
It appears there was a mistake in one of the code samples in the documentation, showing "MicrosoftSQL" as the `storageProvider/type` value in host.json when it should actually be "mssql". This commit fixes the doc mistake.


davidmrdavid (approved):
LGTM!

Issue #178, opened 2023-07-14
Terminate Pending orchestration


jundayin:
Hi, echoing the thread https://github.com/Azure/durabletask/issues/662 in this repo. 

I believe SQL provider works similar to AzureStorageOrchestrationService, which only terminate running orchestration. In our code, we schedule one orchestration which will be triggered in a month. Before it is executed, we also allow user to terminate the orchestration.

However, those scheduled orchestration will be not terminated right away but until the timer expires. Since this is a rather active task for customer, it generates 3K rows in NewEvents table within 3 weeks. 
![Screenshot 2023-07-13 185345](https://github.com/microsoft/durabletask-mssql/assets/74275966/2b290648-7f0f-4439-a4aa-2ba86523c2cb)

Is this a provider behavior or the core behavior? It doesn't seem necessary to terminate the orchestration only when it starts running. The orchestration instance already presents in the db across the board. If it is provider behavior then I think we can have one improvement of terminate both pending and running orchestration in order to reduce the load in NewEvents table

jundayin:
Actually I think this is more a bug instead of expected behavior.

I just glance there are a lot of exceptions log from the pod of our services.
```
Backing off for 1 seconds until 5 successful operations
2023-07-19 00:34:41.8969|ERROR|DurableTask.Core|TaskOrchestrationDispatcher-c6b2e24a2c884db5b6ee1fe277fcfae1-0: Unhandled exception with work item 'aaf0757b-5c35-46e5-9d8b-bc8406ee0289': System.NullReferenceException: Object reference not set to an instance of an object.
   at DurableTask.SqlServer.SqlOrchestrationService.CompleteTaskOrchestrationWorkItemAsync(TaskOrchestrationWorkItem workItem, OrchestrationRuntimeState newRuntimeState, IList`1 outboundMessages, IList`1 orchestratorMessages, IList`1 timerMessages, TaskMessage continuedAsNewMessage, OrchestrationState orchestrationState) in /_/src/DurableTask.SqlServer/SqlOrchestrationService.cs:line 311
   at DurableTask.Core.TaskOrchestrationDispatcher.OnProcessWorkItemAsync(TaskOrchestrationWorkItem workItem) in /_/src/DurableTask.Core/TaskOrchestrationDispatcher.cs:line 548
[...]

PR #276, opened 2024-10-11
Update terminate logic to handle scheduled orchestrations

cgillum:
Fixes https://github.com/microsoft/durabletask-mssql/issues/178

This updates the SQL-based terminate logic to handle the case where users try to terminate scheduled orchestrations. We can't use the previous approach of sending an `ExecutionTerminated` message because DTFx doesn't know how to handle the case where this message is received and there is no orchestration state (you end up with a repeating NullReferenceException until the scheduled orchestration actually starts).

This PR also includes an integration test for this scenario.

### DIFF
--- a/src/DurableTask.SqlServer/DTUtils.cs
+++ b/src/DurableTask.SqlServer/DTUtils.cs
@@ -196,6 +196,7 @@ namespace DurableTask.SqlServer
             return historyEvent.EventType switch
             {
                 EventType.ExecutionStarted => ((ExecutionStartedEvent)historyEvent).Version,
+                EventType.SubOrchestrationInstanceCreated => ((SubOrchestrationInstanceCreatedEvent)historyEvent).Version,
                 EventType.TaskScheduled => ((TaskScheduledEvent)historyEvent).Version,
                 _ => null,
             };
--- a/src/DurableTask.SqlServer/Scripts/logic.sql
+++ b/src/DurableTask.SqlServer/Scripts/logic.sql
@@ -469,45 +469,70 @@ BEGIN
 
     DECLARE @TaskHub varchar(50) = __SchemaNamePlaceholder__.CurrentTaskHub()
 
-    DECLARE @existingStatus varchar(30) = (
-        SELECT TOP 1 existing.[RuntimeStatus]
-        FROM Instances existing WITH (HOLDLOCK)
-        WHERE [TaskHub] = @TaskHub AND [InstanceID] = @InstanceID
-    )
+    DECLARE @existingStatus varchar(30)
+    DECLARE @existingLockExpiration datetime2(7)
+
+    -- Get the status of an existing orchestration
+    SELECT TOP 1
+        @existingStatus = existing.[RuntimeStatus],
+        @existingLockExpiration = existing.[LockExpiration]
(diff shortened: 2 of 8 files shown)


src/DurableTask.SqlServer/DTUtils.cs
```diff
@@ -196,6 +196,7 @@ public static bool HasPayload(HistoryEvent e)
             return historyEvent.EventType switch
             {
                 EventType.ExecutionStarted => ((ExecutionStartedEvent)historyEvent).Version,
+                EventType.SubOrchestrationInstanceCreated => ((SubOrchestrationInstanceCreatedEvent)historyEvent).Version,
```
cgillum, line 199: This is an unrelated fix that I happened to notice independently. I'm just adding it to this PR as a convenience since it's a small change.

src/DurableTask.SqlServer/Scripts/logic.sql
```diff
@@ -469,45 +469,70 @@ BEGIN
...
+        -- Create a payload to store the reason, if any
+        DECLARE @PayloadID uniqueidentifier = NULL
+        IF @Reason IS NOT NULL
         BEGIN
-            -- Payloads are stored separately from the events
-            DECLARE @PayloadID uniqueidentifier = NULL
-            IF @Reason IS NOT NULL
+            -- Note that we don't use the Reason column for the Reason with terminate events
+            SET @PayloadID = NEWID()
+            INSERT INTO Payloads ([TaskHub], [InstanceID], [PayloadID], [Text])
```
andystaples, line 498: The comment in line#468 states that the table order for this sproc should be Instances --> (NewEvents --> Payloads --> NewEvents) but this PR changes the table order to Instances --> (Payloads --> Instances --> NewEvents). 
Does this present a concern for deadlocks, and if not, should we update the comment to reflect the new ordering?

src/DurableTask.SqlServer/Scripts/logic.sql
```diff
@@ -469,45 +469,70 @@ BEGIN
...
+            -- any pending messages
+            UPDATE Instances SET
+                [RuntimeStatus] = 'Terminated',
+                [LastUpdatedTime] = @now,
+                [CompletedTime] = @now,
+                [OutputPayloadID] = @PayloadID,
+                [LockExpiration] = NULL -- release the lock, if any
+            WHERE [TaskHub] = @TaskHub AND [InstanceID] = @InstanceID
+
+            DELETE FROM NewEvents WHERE [TaskHub] = @TaskHub AND [InstanceID] = @InstanceID
```
andystaples, line 515: As it is, the only time we delete from NewEvents is during PurgeInstanceStateByID. Is it best to delete these, or is there a way we can dequeue and preserve them?

src/DurableTask.SqlServer/Scripts/logic.sql
cgillum: Hmm...good observation. Let me take a closer look at this.

lilyjma:
@cgillum /@bachuv - Just bringing this up on your radar again in case this is a small item that can be wrapped up quickly. I think two customers have reported this issue.

PR #277, opened 2024-10-16
Rename source artifacts pipeline for publishing pipeline

davidmrdavid:
I made a copy-pasting error in the PR that introduced the publishing pipeline - it was grabbing the nupkgs from the durabletask-dotnet.official pipeline, not durabletask-mssql.official. This PR fixes that.


nytian (approved):
Thanks for the PR!

PR #281, opened 2024-11-20
Update Durabletask Dependencies for v1.5.0 Release

nytian:
As titled. This PR updates the below dependencies to start the durabletask mssql backend v1.5.0 release:

Microsoft.Azure.WebJobs.Extensions.DurableTask 2.13.7 -> 3.0.0
Microsoft.Azure.DurableTask.Core 2.*-> 3.*

### DIFF
--- a/src/DurableTask.SqlServer.AzureFunctions/DurableTask.SqlServer.AzureFunctions.csproj
+++ b/src/DurableTask.SqlServer.AzureFunctions/DurableTask.SqlServer.AzureFunctions.csproj
@@ -4,7 +4,7 @@
   <Import Project="../common.props" />
   
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
+    <TargetFrameworks>net6.0</TargetFrameworks>
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(TargetFramework)' == 'net6.0'">
@@ -20,7 +20,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="2.13.7" />
+    <PackageReference Include="Microsoft.Azure.WebJobs.Extensions.DurableTask" Version="3.0.0" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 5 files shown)


src/DurableTask.SqlServer.AzureFunctions/DurableTask.SqlServer.AzureFunctions.csproj
```diff
@@ -4,7 +4,7 @@
   <Import Project="../common.props" />
   
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net6.0</TargetFrameworks>
+    <TargetFrameworks>net6.0</TargetFrameworks>
```
nytian, line 7: I need to update the TFM here to successfully build, as the new Webjobs.Extensions.DurableTask v3 only supports .NET 6. However, I'm wondering if we consider updating TFM for mssql here.

cgillum: I'm fine with this. 👍

cgillum (approved):

src/common.props
```diff
@@ -15,8 +15,8 @@
 
   <!-- Version settings: https://andrewlock.net/version-vs-versionsuffix-vs-packageversion-what-do-they-all-mean/ -->
   <PropertyGroup>
-    <MajorVersion>1</MajorVersion>
-    <MinorVersion>4</MinorVersion>
+    <MajorVersion>2</MajorVersion>
+    <MinorVersion>0</MinorVersion>
```
nytian, line 19: @cgillum , I’m wondering if it might make more sense to consider a major version upgrade instead of a minor one, considering we have major version upgrades for both DurableTask.Core and WebJobs.Extensions.DurableTask.

I have a similar PR for Netheriteb https://github.com/microsoft/durabletask-netherite/pull/437, and Sebastian and David both prefer a major version release. Thus, I think I should apply this to mssql  as well.  Let me know your thoughts on this. Thanks!

cgillum: I think I'd rather not make a major version change for MSSQL. There shouldn't be any breaking dependency or behavior changes, so I feel like doing a major version increment is a bit misleading. Semantic versioning AFAIK doesn't make any recommendation about how to handle major version changes for dependencies, so I think it's up to the library owner's discretion. More discussion on this [here](https://softwareengineering.stackexchange.com/a/168661). I'm okay if MSSQL uses a different versioning strategy than Netherite.

PR #283, opened 2025-01-22
Move to central package management

cgillum:
In order to make it easier to keep up with vulnerability management in this repo, we're changing it to use central package management.

While I tried not to make any actual dependency changes, there were some changes that couldn't be avoided due to inconsistencies in certain referenced packages across different projects. The only intentional change as part of this PR is the update of Microsoft.Data.SqlClient from 3.1.5 to 3.1.7. Additional updates will come as part of a later PR.


bachuv (approved):
LGTM!

PR #285, opened 2025-02-10
Use Singleton Metrics Provider for Listener

nytian:
- Use Singleton metrics provider instead of  IScaleMonitor and ITargetScaler
- Update SqlScaleMonitor and SqlTargeSclaer to be per function id.
   Similar changes at AzureStorage backend https://github.com/Azure/azure-functions-durable-extension/commit/c5d8e04d24d0343438a89c616f890a9a2c20cf48#diff-09935cd15a9926e82e0d7ff3df8d480813cabd51b8e3a483985ec5a0da3035cd
- Update Microsoft.Azure.WebJobs.Extensions.DurableTask from 3.0.2 -> 3.0.3 at Directory.Packages.props


src/DurableTask.SqlServer.AzureFunctions/SqlDurabilityProvider.cs
```diff
@@ -200,13 +197,12 @@ public override bool TryGetScaleMonitor(
...
-            if (this.scaleMonitor == null)
+            if (this.singletonSqlMetricsProvider == null)
             {
-                var sqlMetricsProvider = new SqlMetricsProvider(this.service);
-                this.scaleMonitor = new SqlScaleMonitor(hubName, sqlMetricsProvider);
+                this.singletonSqlMetricsProvider = new SqlMetricsProvider(this.service);
             }
-
-            scaleMonitor = this.scaleMonitor;
+            scaleMonitor = new SqlScaleMonitor(hubName, this.singletonSqlMetricsProvider);
```
cgillum, line 204: Based on the changes Alexey made in the Azure Storage repo, I think we also need to include the `functionId` value in the monitor that we create. That way each monitor has a unique name/ID.

src/DurableTask.SqlServer.AzureFunctions/SqlDurabilityProvider.cs
```diff
@@ -218,13 +214,12 @@ public override bool TryGetTargetScaler(
...
-            if (this.targetScaler == null)
+            if (this.singletonSqlMetricsProvider == null)
             {
-                var sqlMetricsProvider = new SqlMetricsProvider(this.service);
-                this.targetScaler = new SqlTargetScaler(hubName, sqlMetricsProvider);
+                this.singletonSqlMetricsProvider = new SqlMetricsProvider(this.service);
             }
 
-            targetScaler = this.targetScaler;
+            targetScaler = new SqlTargetScaler(hubName, this.singletonSqlMetricsProvider);
```
cgillum, line 222: Same comment about using the function ID to ensure uniqueness in the target scaler.

cgillum (commented):
I think we need a couple more changes.

src/DurableTask.SqlServer.AzureFunctions/SqlTargetScaler.cs
```diff
@@ -12,14 +12,13 @@ public class SqlTargetScaler : ITargetScaler
...
         readonly SqlMetricsProvider sqlMetricsProvider;
 
-        public SqlTargetScaler(string taskHubName, SqlMetricsProvider sqlMetricsProvider)
+        public SqlTargetScaler(string functionId, SqlMetricsProvider sqlMetricsProvider)
         {
             this.sqlMetricsProvider = sqlMetricsProvider;
 
             // Scalers in Durable Functions are shared for all functions in the same task hub.
             // So instead of using a function ID, we use the task hub name as the basis for the descriptor ID.
-            string id = $"DurableTask-SqlServer:{taskHubName ?? "default"}";
```
nytian, line 21: Seems like we don't need this, but only the function id to create the target scaler descriptor. How we created this at AzureStorage ([here](https://github.com/Azure/azure-functions-durable-extension/blob/dev/src/WebJobs.Extensions.DurableTask/Listener/DurableTaskTargetScaler.cs#L32) )

src/DurableTask.SqlServer.AzureFunctions/SqlScaleMonitor.cs
```diff
@@ -23,14 +23,13 @@ class SqlScaleMonitor : IScaleMonitor<SqlScaleMetric>
 
         int? previousWorkerCount = -1;
 
-        public SqlScaleMonitor(string taskHubName, SqlMetricsProvider sqlMetricsProvider)
+        public SqlScaleMonitor(string functionId, string taskHubName, SqlMetricsProvider sqlMetricsProvider)
         {
-            // Scalers in Durable Functions are shared for all functions in the same task hub.
-            // So instead of using a function ID, we use the task hub name as the basis for the descriptor ID.
+            // Scalers in Durable Functions is per function ids. And scalers share the same sqlMetricsProvider in the same taskhub. 
             string id = $"DurableTask-SqlServer:{taskHubName ?? "default"}";
```
alrod, line 29: string id = $"{functionId}-DurableTask-SqlServer:{taskHubName ?? "default"}";
```suggestion
            string id = $"DurableTask-SqlServer:{taskHubName ?? "default"}";
```

nytian: Thanks! does this string id needs to add ToLower(CultureInfo.InvariantCulture) to be lower case?

src/DurableTask.SqlServer.AzureFunctions/SqlDurabilityProvider.cs
```diff
@@ -200,13 +197,12 @@ public override bool TryGetScaleMonitor(
             string storageConnectionString,
             out IScaleMonitor scaleMonitor)
         {
-            if (this.scaleMonitor == null)
+            if (this.singletonSqlMetricsProvider == null)
             {
-                var sqlMetricsProvider = new SqlMetricsProvider(this.service);
-                this.scaleMonitor = new SqlScaleMonitor(hubName, sqlMetricsProvider);
+                this.singletonSqlMetricsProvider = new SqlMetricsProvider(this.service);
```
cgillum, line 202: One last thing. I recall the final fix that Alexey made to the Azure Storage metrics provider was to throttle it so that we only fetch metrics once every few seconds or so. This is how we ensured that we don't have too many storage transactions, even if the app has a lot of triggers. I think it will be important to do that for the `SqlMetricsProvider` too so that we don't overwhelm the customer's database with expensive SQL queries.

src/DurableTask.SqlServer.AzureFunctions/SqlMetricsProvider.cs
```diff
@@ -3,12 +3,15 @@
...
 {
+    using System;
     using System.Threading;
     using System.Threading.Tasks;
 
     public class SqlMetricsProvider
     {
         readonly SqlOrchestrationService service;
+        DateTime metricsTimeStamp = DateTime.MinValue;
+        SqlScaleMetric? metrics;
```
nytian, line 14: Added this cache. If the target scaler calls `GetMetricsAsync` within 5 seconds, return the last result to make the target scaler remains the same decision as last time.

cgillum (approved):
Looks good!

PR #295, opened 2025-11-10
Add support for Durable Entities for functions ooproc languages

andystaples:
## Summary

Adds Durable Entities support to the SQL durability provider for Azure Functions out-of-process worker languages, including .NET isolated, Java, and custom workers. The SQL backend now implements `IEntityOrchestrationService`, supports entity management queries, and can dispatch entity and orchestration work on separate queues when required by the worker runtime.

## Changes

- Implements `IEntityOrchestrationService` in the SQL orchestration service and exposes entity backend capabilities and queries.
- Adds entity APIs for retrieving and querying entity metadata and state, removing empty entities, and releasing orphaned locks, with paged continuation-token support.
- Adds configurable entity behavior through `SqlOrchestrationServiceSettings`:
  - Maximum concurrent entity work items
  - Maximum entity operation batch size
  - Entity message reorder window
  - Shared or separate entity work-item queues
- Extends the SQL work-item locking procedure to filter orchestration and entity instances, while retaining shared-queue behavior by default for compatibility.
- Adds dedicated entity and orchestration lock paths and supports switching between shared and separate queue modes without losing or double-locking pending work.
- Automatically enables separate entity dispatch for .NET isolated, Java, and custom Azure Functions workers, and wires Durable Functions entity concurrency and batching options into the SQL backend.
- Updates `Microsoft.Azure.WebJobs.Extensions.DurableTask` from `3.0.3` to `3.10.0`.
- Moves `DurableTask.SqlServer.AzureFunctions`, its test project, and the performance test project from `net6.0` to `net8.0`. This does not introduce a major package version bump, consistent with other Durable Task packages.
- Updates GitHub Actions and Azure Pipelines build/release configuration to install the .NET 8 SDK alongside .NET 6, and updates `setup-dotnet` usage to v4.
- Documents Durable Entities and the `net8.0` target change in the changelog.

## Test coverage

- Verifies Azure Functions provider settings propagate separate entity queue dispatch to the backend.
- Verifies existing work is partitioned correctly when switching to separate queues.
- Verifies shared queues continue to drain entity work when switching back.
- Verifies shared and separate workers do not double-lock entity work.
- Verifies entity metadata is returned in the shape expected by Durable Task.


OmnipotentOwl:
I was able to fork the repo with this branch in it and run through the testing to debug what might be occurring with the failed tests in the pipeline. From what I could see when running in Visual Studio locally, the current tests primarily cover the in-process model and don't seem to test the out-of-process model for its code flows.

I went ahead and, for testing purposes, rebuilt the PerformanceTests testing function app to migrate it over to the Isolated hosting model (out-of-process) and was able to successfully run all of the performance test flows using localdb on my machine. The entities test battery I ran was for StartManyEntities with 1000 entities and 1000 messages to each entity. From that testing, I found no issues so far and was able to pack each of the NuGet packages into a local feed and run through a more operational test suite using it in a development application. The development application was using Azure SQL for the database hosting, and the in-place upgrade of the SQL schema went without any issue. Additionally, in the development application, I was able to send 5000 entity creation and signaling events as well as calls from orchestrators through without any issue, so in terms of out-of-process functionality, everything looks solid.

With respect to the in-process tests that are failing, my observations were that the function app/tests would create the entities, but they would always stay stuck in a 'Pending' state and never get started. I wasn't able to get to the root cause of that in my time digging through the code.

@cgillum or @andystaples Do you want me to contribute the PerformanceTests Isolated Function App back to the repo so you have a test bed for Isolated to ensure changes work properly? I did have some headaches with having to pack the NuGets first before that function app could build or run, so there might need to be some guidance in the repo when trying to run that project.

OmnipotentOwl:
@cgillum, any update with this PR so we can close this and get it merged? Ive been working with a customer that is having to custom build this and fold in the upstream changes and every time this repo releases a new version it breaks all of their apps that depend on the custom built version while they work around the lack of this core functionality on out-of-process functions.

With in-process being deprecated in November and no alternative solution available for Kubernetes based customers this is putting customers in difficult spot to keep using Durable Functions itself. https://azure.microsoft.com/en-us/updates?id=retirement-support-for-the-inprocess-model-for-net-apps-in-azure-functions-ends-10-november-2026

cgillum:
Agreed that we need to prioritize this since in-process is going out of support. FYI @torosent

OmnipotentOwl:
Do you want me to PR in the PerformanceTests I rebuilt for the out of process function App Library?

dreeko:
We’ve been testing this in our self-hosted deployment and Durable Entities have been working well for us in the .NET isolated model. We’re currently carrying the PR as a fork patch, so each upstream release creates maintenance work.

src/DurableTask.SqlServer/EntitySqlBackendQueries.cs
```diff
@@ -0,0 +1,217 @@
...
+
+                        if (request.RemoveEmptyEntities)
+                        {
+                            bool isEmptyEntity = !status.EntityExists && status.LockedBy == null && status.BacklogQueueSize == 0;
+                            bool safeToRemoveWithoutBreakingMessageSorterLogic =
+                                now - state.LastUpdatedTime > this.orchestrationService.EntityBackendProperties.EntityMessageReorderWindow;
+                            if (isEmptyEntity && safeToRemoveWithoutBreakingMessageSorterLogic)
+                            {
+                                emptyEntityIds.Add(state.OrchestrationInstance.InstanceId);
+                                emptyEntitiesRemoved++;
```
bachuv: Is `EmptyEntitiesRemoved` double-counted? Each empty entity is counted here when it's added to `emptyEntityIds`, and then counted again on line 164 via the return value of `PurgeOrchestrationHistoryAsync` (which returns the number of instances actually purged, Task<int>).

src/DurableTask.SqlServer/SqlOrchestrationServiceSettings.cs
```diff
@@ -141,6 +148,35 @@ public SqlOrchestrationServiceSettings(string connectionString, string? taskHubN
...
+        /// </summary>
+        /// <remarks>
+        /// Limiting the batch size can help to avoid timeouts in execution environments that impose time limitations on work items.
+        /// If set to 1, batching is disabled, and each operation executes as a separate work item.
+        /// </remarks>
+        /// <value>
+        /// A positive integer, or null.
+        /// </value>
+        [JsonProperty("maxEntityOperationBatchSize")]
+        public int? MaxEntityOperationBatchSize { get; set; } = null;
```
bachuv, line 163: Does Azure Storage also default to null (no limit)?

src/DurableTask.SqlServer.AzureFunctions/DurableTask.SqlServer.AzureFunctions.csproj
```diff
@@ -4,10 +4,10 @@
   <Import Project="../common.props" />
   
   <PropertyGroup>
-    <TargetFrameworks>net6.0</TargetFrameworks>
+    <TargetFrameworks>net8.0</TargetFrameworks>
```
bachuv, line 7: Will we release a new major version for this .NET change? Either way, let's document this in the PR description, the CHANGELOG.md, and release notes.

src/DurableTask.SqlServer/Scripts/logic.sql
```diff
@@ -672,7 +674,11 @@ BEGIN
     WHERE
         I.TaskHub = @TaskHub AND
 	    (I.[LockExpiration] IS NULL OR I.[LockExpiration] < @now) AND
-        (E.[VisibleTime] IS NULL OR E.[VisibleTime] < @now)
+        (E.[VisibleTime] IS NULL OR E.[VisibleTime] < @now) AND
+        (@OrchestrationType IS NULL OR
+            (@OrchestrationType = 0 AND I.[InstanceID] NOT LIKE '@%@%') OR
+            (@OrchestrationType = 1 AND I.[InstanceID] LIKE '@%@%')
```
bachuv, line 680: Should we match Azure Storage on how to identify an entity by checking for one "@"?  https://github.com/Azure/durabletask/blob/f92d1584f8319327d22ff27bc2bb5f013fc3ffdf/src/DurableTask.Core/Common/Entities.cs#L30

src/DurableTask.SqlServer/EntitySqlBackendQueries.cs
andystaples: Good catch, thanks! This was indeed double-counted - once when added to emptyEntityIds and again from the return value of PurgeOrchestrationHistoryAsync. Fixed by removing the redundant increment.

src/DurableTask.SqlServer/SqlOrchestrationServiceSettings.cs
andystaples: Yep, checked upstream - AzureStorageOrchestrationServiceSettings.MaxEntityOperationBatchSize also defaults to null (no limit), so this is consistent.

src/DurableTask.SqlServer.AzureFunctions/DurableTask.SqlServer.AzureFunctions.csproj
andystaples: No major version bump - we didn't bump major for any of the other Durable Task packages that added net8.0 targets either, so keeping this consistent. Added a note about the net8.0 addition to CHANGELOG.md under Unreleased (1.7.0 is already released, so this will ship in the next release).

src/DurableTask.SqlServer/Scripts/logic.sql
andystaples: Good point - Core's canonical Entities.IsEntityInstance only checks for a single leading '@', while this checks for two '@'s (matching EntityId.ToString()'s '@Name@Key' format). Today all auto-started entity orchestrations from this provider follow that two-'@' format, so the check is correct in practice, but it's not strictly aligned with the Core convention. Leaving as-is for now since changing it would affect dispatch semantics, but flagging for follow-up if we ever need to recognize instance IDs from other backends/producers that only use a single '@'.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+### Updates
+
+* Add support for Durable Entities, including a new `net8.0` target framework for `DurableTask.SqlServer.AzureFunctions` (no major version bump, consistent with other Durable Task packages) ([#295](https://github.com/microsoft/durabletask-mssql/pull/295))
```
bachuv: nit: can we remove `(no major version bump, consistent with other Durable Task packages)`?

> Add support for Durable Entities, including a new `net8.0` target framework for `DurableTask.SqlServer.AzureFunctions` ([#295](https://github.com/microsoft/durabletask-mssql/pull/295))

andystaples: Done

bachuv (approved):

andystaples:
Hi @OmnipotentOwl , @dreeko 
Thank you for your patience on this. The feature has now been merged and is released in 1.8.0.

dreeko:
Brilliant,  great work folks! 🎖

PR #299, opened 2026-02-25
Implement orchestration tags

fanyirobin:


fanyirobin:
@microsoft-github-policy-service agree company="Microsoft"

src/DurableTask.SqlServer/SqlUtils.cs
```diff
@@ -478,6 +479,51 @@ internal static SqlString GetTraceContext(HistoryEvent e)
...
+        }
+
+        internal static void AddTagsParameter(
+            this SqlParameterCollection parameters,
+            IDictionary<string, string>? tags)
+        {
+            string? json = tags != null && tags.Count > 0
+                ? DTUtils.SerializeToJson(tags)
+                : null;
+            parameters.Add("@Tags", SqlDbType.VarChar, 8000).Value = (object?)json ?? DBNull.Value;
```
cgillum: Given we have max length of 8000 characters, I think we need to validate the size of the tags payload to make sure it fits so that we don't run into data corruption issues. I suggest we do this at the time of orchestration creation so that we can fail early if the tags payload is too large.

fanyirobin: makes sense to me. Added size validation in AddParameter helper and it is called at the beginning of the CreateTaskOrchestrationAsync function.

src/DurableTask.SqlServer/SqlUtils.cs
```diff
@@ -478,6 +479,51 @@ internal static SqlString GetTraceContext(HistoryEvent e)
...
         }
 
+        internal static IDictionary<string, string>? GetTags(DbDataReader reader)
+        {
+            int ordinal;
+            try
+            {
+                ordinal = reader.GetOrdinal("Tags");
+            }
+            catch (IndexOutOfRangeException)
```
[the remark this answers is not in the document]
fanyirobin: it shouldn't be necessary, since we have schema migration changes in the same PR. removed in latest commit

src/DurableTask.SqlServer/SqlUtils.cs
```diff
@@ -511,15 +511,18 @@ internal static SqlString GetTraceContext(HistoryEvent e)
             }
         }
 
-        internal static SqlString GetTagsJson(HistoryEvent e)
+        internal static SqlString GetTagsJson(HistoryEvent e, LogHelper logHelper)
         {
             if (e is ExecutionStartedEvent startedEvent && startedEvent.Tags != null && startedEvent.Tags.Count > 0)
             {
                 string json = DTUtils.SerializeToJson(startedEvent.Tags);
                 if (json.Length > MaxTagsPayloadSize)
```
cgillum: This isn't the right way to check the payload size. We need to check the byte length, not the string length. Use this instead:
```csharp
int utf8Bytes = Encoding.UTF8.GetByteCount(json);
if (utf8Bytes > MaxTagsPayloadSize)
{
   throw new ArgumentException(
       $"The serialized tags payload is {utf8Bytes} bytes, which exceeds the maximum allowed size of {MaxTagsPayloadSize} bytes.");
}
```
Same for any other place where we're doing a size check.

cgillum (approved):
LGTM! Thanks for your patience on this PR.

GouthamiSivakoti:
can we get tags in OrchestrationContext ?

PR #308, opened 2026-04-21
enable SqlOrchestrationService.PurgeInstanceStateAsync to delete more than 1000 instances

### LINKED ISSUE #297: PurgeAllInstancesAsync - Durable SQL Provider - Deletes only 1000 instances maximum
We see a behavior, where, only a 1000 instances have been deleted though there are many eligible instances are available in the 'Instances' table.

**Is this behavior a default one?** 
Is there a way to increase this limit of 1000?

Our intention is to run the PurgeAll once every day but with this restriction we left with millions of records in our sql db, indirectly hindering our running durable orchestrators performance.
```

[Function("PurgeDurableFunctionsHistory")]
public async Task RunAsync([TimerTrigger("%PurgeDurableFunctionsHistory:Schedule%")] TimerInfo myTimer, [DurableClient] DurableTaskClient client, CancellationToken cancellationToken)
{
PurgeResult result = null;
if (!int.TryParse(Environment.GetEnvironmentVariable("PurgeDurableFunctionsHistoryDays"), out int deletePurgeHistoryDays))
{
deletePurgeHistoryDays = 30;
}

var orchStatus = new List<OrchestrationRuntimeStatus>
{
	OrchestrationRuntimeStatus.Completed,
};
var purgeFilter = new PurgeInstancesFilter(
	DateTime.SpecifyKind(DateTime.MinValue, DateTimeKind.Utc),
	DateTime.SpecifyKind(DateTime.UtcNow.AddDays(deletePurgeHistoryDays * -1), DateTimeKind.Utc),
	orchStatus);

result = await client.PurgeAllInstancesAsync(purgeFilter, cancellationToken).ConfigureAwait(false);
}
```
NuGet used:

```
<PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.DurableTask" Version="1.1.4" />
<PackageReference Include="Microsoft.Azure.Functions.Worker.Extensions.DurableTask.SqlServer" Version="1.3.0" />		
<PackageReference Include="Microsoft.Azure.Functions.Worker" Version="1.22.0" />
```

usemam:
Supposed to fix https://github.com/microsoft/durabletask-mssql/issues/297.

### Summary of changes
- New `PurgeInstanceStateByFilter` stored procedure (`logic.sql`): Accepts `@CreatedTimeFrom`, `@CreatedTimeTo`, and `@RuntimeStatusFilter` parameters, selects all matching instance IDs server-side, and delegates deletion to the existing `PurgeInstanceStateByID` sproc. This follows the same pattern as the existing `PurgeInstanceStateByTime` procedure.
- Rewritten `PurgeInstanceStateAsync(PurgeInstanceFilter)` (`SqlOrchestrationService.cs`): Replaced the two-step query-then-delete approach (limited to 1000) with a single call to the new stored procedure, removing the artificial limit entirely.
- Integration test (`DataRetentionTests.cs`): Added `PurgesMoreThanOneThousandInstances` test that creates 1100 completed orchestrations and verifies all are purged in a single `PurgeInstanceStateAsync` call.

### DIFF
--- a/src/DurableTask.SqlServer/Scripts/drop-schema.sql
+++ b/src/DurableTask.SqlServer/Scripts/drop-schema.sql
@@ -19,6 +19,7 @@ DROP PROCEDURE IF EXISTS __SchemaNamePlaceholder__.SetGlobalSetting
 DROP PROCEDURE IF EXISTS __SchemaNamePlaceholder__.TerminateInstance
 DROP PROCEDURE IF EXISTS __SchemaNamePlaceholder__.PurgeInstanceStateByID
 DROP PROCEDURE IF EXISTS __SchemaNamePlaceholder__.PurgeInstanceStateByTime
+DROP PROCEDURE IF EXISTS __SchemaNamePlaceholder__.PurgeInstanceStateByFilter
 
 -- Private sprocs
 DROP PROCEDURE IF EXISTS __SchemaNamePlaceholder__._AddOrchestrationEvents
--- a/src/DurableTask.SqlServer/Scripts/drop-schema.sql
+++ b/src/DurableTask.SqlServer/Scripts/drop-schema.sql
@@ -19,7 +19,6 @@ DROP PROCEDURE IF EXISTS __SchemaNamePlaceholder__.SetGlobalSetting
 DROP PROCEDURE IF EXISTS __SchemaNamePlaceholder__.TerminateInstance
 DROP PROCEDURE IF EXISTS __SchemaNamePlaceholder__.PurgeInstanceStateByID
 DROP PROCEDURE IF EXISTS __SchemaNamePlaceholder__.PurgeInstanceStateByTime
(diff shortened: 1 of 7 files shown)


cgillum (commented):
Thank you for this PR! One concern I have is that the list of instances to be purged could grow extremely large, in which case this transaction may never be able to succeed due to memory constraints on the database. Should we consider a design where we keep the original implementation (which includes the 1000 `PageSize` limit) but run it in a loop so that we purge in batches of 1000, thus capping the maximum size of the database transaction?

usemam:
Sounds good. I was initially apprehensive about iterative approach, but in the case of extremely large pool of instances to be purged I'd agree that it'll perform better.
Pushed my latest changes - please take a look when possible.
Thank you!

cgillum (approved):
LGTM, thanks!

cgillum:
I just now triggered the CI to validate the changes.

Issue #56, opened 2021-11-20
Extended Sessions


cgillum:
The MSSQL backend doesn't yet support [extended sessions](https://docs.microsoft.com/azure/azure-functions/durable/durable-functions-perf-and-scale#extended-sessions), which could provide significant performance benefits in many scenarios. This issue tracks adding extended session support.

usemam:
Took a stab at this here - https://github.com/microsoft/durabletask-mssql/pull/309. @cgillum , I would appreciate if you can take a look. Thank you!

PR #309, opened 2026-05-04
Extended session support

usemam:
### Summary
Keep orchestration in memory until it's idle for a configured timeout.
When extended sessions enabled, check-pointing the orchestration instance does not reset `Instances.LockedBy`/`LockExpiration` values. A `SqlOrchestrationSession` instance is provided to the dispatcher, through which new messages for the orchestration instance are fetched.

### Links
- [Existing doc](https://learn.microsoft.com/en-us/azure/azure-functions/durable-functions/durable-functions-perf-and-scale#caching-by-storage-provider)
- https://github.com/microsoft/durabletask-mssql/issues/56


usemam:
@cgillum sorry to bother - could you please provide your feedback? 🙏

usemam:
@cgillum , apologies for pinging - could you please review this one? 🙏

cgillum:
Hey @usemam, sorry for missing your pings! Adding @sophiatev to take a look at this PR as she has the most context on extended sessions these days.

sophiatev:
Hey @usemam can you fix the merge conflicts, then I'll take a look :)

usemam:
@sophiatev apologies about the delay. Conflicts are fixed now. Please take a look. Thank you!

andystaples (dismissed):
Thanks for putting this together. The overall approach—implementing Core's `IOrchestrationSession`, defaulting the feature off, and retaining the SQL instance lock across checkpoints—is the right model for this backend. I found the following blockers that need to be addressed before merge:

1. **The PR does not compile.** `SqlOrchestrationSession` implements `IOrchestrationSession` but does not implement `EndSessionAsync()`. The current head fails with CS0535 for netstandard2.0, net8.0, and net10.0. A no-op implementation is appropriate if lock release remains in `ReleaseTaskOrchestrationWorkItemAsync`.

2. **Lease ownership is not generation-safe.** Every session receives the process-wide `lockedByValue`, while checkpoint, fetch, renew, and release identify ownership using that reusable value. If a lease expires and the same process reacquires the instance, the old session can still checkpoint or clear the replacement session's lease. Also, the checkpoint guard checks `LockedBy` but not whether `LockExpiration` is still valid, and `_RenewOrchestrationLocks` has no owner guard. Please use a unique token per lock acquisition and validate that token plus an unexpired lease in checkpoint/fetch/renew/release; lock-loss should surface as `SessionAbortedException`.

3. **The retained `EventPayloadMap` collides across ContinueAsNew generations.** Value-tracked payloads are keyed only by `(EventType, TaskEventId)` and inserted with `Dictionary.Add`. Activity and sub-orchestration IDs restart after ContinueAsNew, so scheduling ID 0 in a later generation can throw before checkpointing. Reset/replace mappings after a successful checkpoint or include execution identity in the key.

4. **`_FetchOrchestrationMessages` returns an undefined event order.** It uses `TOP (@BatchSize)` without `ORDER BY`. Core only ensures that `ExecutionStarted` is first; it does not restore order among external events or task responses. Please add `ORDER BY N.SequenceNumber ASC`.

5. **All sessions share the service's stateful `BackoffPollingHelper`.** The mutable interval, exponent, `Random`, and auto-reset event are shared with the main orchestration poller and every concurrent session. Session polling therefore changes global polling latency, and reset signals can wake an unrelated waiter. Each session should own an independent helper and reset it after finding work.

6. **Azure Functions timeout mapping is inconsistent with Azure Storage.** `SqlDurabilityOptions` only assigns `ExtendedSessionIdleTimeout` when the common option is greater than zero. Explicit zero/negative values silently retain SQL's 30-second default, while Azure Storage clamps the same option to zero; this can also disagree with the isolated worker cache configuration. Please align the mapping (for example, clamp to zero) or reject invalid values consistently.

Please add repository-local coverage for:

- the Core interface/build contract;
- lease expiry followed by same-process reacquisition, proving stale checkpoint, renew, fetch, and release calls are rejected;
- order-sensitive external events or responses across a session;
- ContinueAsNew followed by activities and sub-orchestrations whose task IDs restart;
- Azure Functions adapter mapping for default, custom, zero, and negative timeout values;
- Durable Entity behavior in both queue modes, or explicitly prevent sessions from being attached to entity work items until supported;
- graceful shutdown/drain and lock release.

The cross-repository package pins, extension bundle publication, public docs/tooling schemas, and rollout telemetry can be handled as maintainer-owned follow-ups rather than requirements for this contribution.

andystaples:
Hi @usemam , thanks for your contribution. Independent of Sophia's review, I had an agent take a look and it produced the findings above. Please take a look and address where applicable, and thanks again!

usemam:
Hi @andystaples,
Thanks for the detailed review. Here’s what changed for each finding mentioned in your comment:

### Fixed 

1. **`EndSessionAsync` (non-compiling / missing).**
2. **Generation-safe lease ownership.** Extended sessions now use a unique per-acquisition token (`appName,pid|guid`) instead of shared process-wide `lockedByValue`. It's generated in `LockNextWorkItemAsync`, passed to `_LockNextOrchestration`, stored on the work item (`ExtendedOrchestrationWorkItem.LockOwnershipToken`), and handed to the `SqlOrchestrationSession`. Every lock-guarded proc now validates owner + unexpired lease:
  - `_RenewOrchestrationLocks` gained `@LockedBy` and throws `50003` when the lease is lost.
  - `_CheckpointOrchestration` validates owner and `LockExpiration > now`, distinguishing `50003` (lock lost) from `50000` (instance missing).
  - `_FetchOrchestrationMessages` / `_ReleaseOrchestrationLock` were already guarded; the unique token makes them generation-safe.
3. **`EventPayloadMap` collision across `ContinueAsNew`.** Added `EventPayloadMap.Clear()` and call it after a successful checkpoint for session work items.
4. **`_FetchOrchestrationMessages` event order.** Added `ORDER BY N.[SequenceNumber] ASC` so the batch is deterministic.
5. **Per-session `BackoffPollingHelper`.**
6. **Azure Functions timeout mapping.**
7. **Added test for lock release on shutdown.**

### Left out of scope: full Durable Entity support

I'd like to keep this out to reduce the scope of current PR. I will follow up with an issue to add a full support under extended sessions for this.

usemam:
Hi @andystaples . Just following up to see if you had a chance to take a look at fixes I've made. Thank you.

usemam:
Hi @sophiatev and @andystaples .
Apologies for pinging you again. Just wanted to check in to see if you had a chance to take look at this PR?

sophiatev:
Hi @usemam sorry for the delay, I'll take a look

src/DurableTask.SqlServer/SqlOrchestrationService.cs
```diff
@@ -178,8 +178,12 @@ public override async Task<TaskOrchestrationWorkItem> LockNextEntityWorkItemAsyn
                 int batchSize = this.settings.WorkItemBatchSize;
                 DateTime lockExpiration = DateTime.UtcNow.Add(this.settings.WorkItemLockTimeout);
 
+                string lockOwnershipToken = this.settings.ExtendedSessionsEnabled
+                    ? $"{this.lockedByValue}|{Guid.NewGuid():N}"
```
sophiatev: Can you explain why we need the randomized GUID in the extended sessions case?

test/DurableTask.SqlServer.AzureFunctions.Tests/ExtendedSessionOptionsTests.cs
```diff
@@ -0,0 +1,62 @@
...
+        [Fact]
+        public void UnconfiguredTimeoutMapsFrameworkDefault()
+        {
+            var extensionOptions = new DurableTaskOptions();
+            int frameworkDefault = extensionOptions.ExtendedSessionIdleTimeoutInSeconds;
+
+            SqlOrchestrationServiceSettings settings = GetSettings(extensionOptions);
+
+            Assert.True(frameworkDefault > 0, "Expected a positive framework default for the unconfigured case.");
+            Assert.Equal(TimeSpan.FromSeconds(Math.Max(0, frameworkDefault)), settings.ExtendedSessionIdleTimeout);
```
sophiatev: Nit: We already asserted the framework default is positive so this Math.Max is unnecessary here

test/DurableTask.SqlServer.Tests/Integration/ExtendedSessionTests.cs
```diff
@@ -0,0 +1,460 @@
...
+                    return tcs.Task;
+                },
+                onEvent: (ctx, name, value) => tcs.TrySetResult(JsonConvert.DeserializeObject<string>(value)));
+
+            await instance.WaitForStart();
+            string heldBy = await this.WaitForLockToBeHeldAsync(instance.InstanceId, TimeSpan.FromSeconds(10));
+
+            // Sanity check: a second worker calling _LockNextOrchestration sees no available work
+            // because the only ready instance is locked by the in-flight session.
+            object available = await SharedTestHelpers.ExecuteSqlAsync(
```
sophiatev: Rather than manually writing SQL here can we actually try to start up another worker and confirm LockNextOrchestration returns nothing?

test/DurableTask.SqlServer.Tests/Integration/ExtendedSessionTests.cs
```diff
@@ -0,0 +1,460 @@
...
+                {
+                    tcs = new TaskCompletionSource<string>();
+                    return tcs.Task;
+                },
+                onEvent: (ctx, name, value) => tcs.TrySetResult(JsonConvert.DeserializeObject<string>(value)));
+
+            await instance.WaitForStart();
+            await this.WaitForLockToBeHeldAsync(instance.InstanceId, TimeSpan.FromSeconds(10));
+
+            // Forcibly clear the lock
```
sophiatev: This should cause a `SessionAbortedException`, right? Can we confirm that in the logs?

test/DurableTask.SqlServer.Tests/Integration/ExtendedSessionTests.cs
```diff
@@ -0,0 +1,460 @@
...
+            // Even the correct owner is rejected once the lease has expired.
+            await SharedTestHelpers.ExecuteSqlAsync(
+                this.output,
+                $"UPDATE dt.[Instances] SET [LockExpiration] = DATEADD(MINUTE, -5, SYSUTCDATETIME()) WHERE [InstanceID] = '{id}'");
+            Assert.Equal(50003, await this.RunGuardedProcAsync(
+                $"EXEC dt._RenewOrchestrationLocks @InstanceID='{id}', @LockExpiration='{future}', @LockedBy='{liveToken}';"));
+
+            // The orchestration still drives to completion after the lease is lost: the stale session
+            // aborts and a fresh lock acquisition takes over.
+            await instance.RaiseEventAsync("Continue", "done");
```
sophiatev: Again can we confirm `SessionAbortedException` is thrown?

src/DurableTask.SqlServer/SqlOrchestrationService.cs
usemam: Please refer to my [comment above](https://github.com/microsoft/durabletask-mssql/pull/309#issuecomment-5471613589) that addresses @andystaples concerns, in particular

> Generation-safe lease ownership. Extended sessions now use a unique per-acquisition token (appName,pid|guid) instead of shared process-wide lockedByValue.

Or do you mean that we don't need to use GUID to make lock token unique within the context of `SqlOrchestrationService` ?

sophiatev: Ah yeah that answers my question. Can you just add a comment explaining why this is necessary (that the same process can stop and start extended sessions on the same instance so we need to distinguish across the sessions)

usemam: Added comment in a2efb6d

test/DurableTask.SqlServer.Tests/Integration/ExtendedSessionTests.cs
usemam: Rewrote this test in a2efb6d

test/DurableTask.SqlServer.Tests/Integration/ExtendedSessionTests.cs
usemam: Added log check in a2efb6d

test/DurableTask.SqlServer.Tests/Integration/ExtendedSessionTests.cs
usemam: Same here - added log check. Please see a2efb6d

sophiatev (approved):

PR #319, opened 2026-07-27
Support separate entity work-item dispatch

andystaples:
## Summary

- implement `IEntityOrchestrationService` in the SQL backend with independent entity and orchestration dequeue paths
- partition entity work at lock time using the existing `@name@key` instance ID convention
- keep the existing combined dequeue path for shared mode and older workers
- make the provider and factory apply `UseSeparateQueueForEntityWorkItems` before `TaskHubWorker` construction
- preserve .NET 6 support while adding a .NET 8 target using Durable Functions extension 3.10.0
- update CI/release SDK installation and advance the package version to 1.8.0 so existing databases reapply the stored procedure update

This supplies the storage-provider behavior required by Azure/azure-functions-durable-extension#3260. Entity and orchestration messages continue to share `NewEvents`; switching modes changes only the atomic lock-time filter, so queued work is not moved or stranded.

## Compatibility

`_LockNextOrchestration` adds an optional `@IsEntity` parameter:

- `NULL`: combined entity/orchestration dispatch used by shared mode and older workers
- `0`: orchestration-only dispatch
- `1`: entity-only dispatch

All three paths lock the same `Instances` row in one transaction. This preserves exactly-once locking during rolling upgrades where shared-mode and separate-mode workers use the same task hub. Events are returned in `SequenceNumber` order so the backend's zero entity reorder window remains valid.

## Validation

- restored and built the complete solution for all target frameworks
- verified provider/factory mode switching and native entity support
- verified false-to-true migration with already-enqueued work
- verified true-to-false migration drains existing entity work through the combined path
- verified concurrent shared and separate workers cannot lock the same entity instance
- verified the final stored procedure against SQL Server 2019

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -20,10 +20,12 @@ jobs:
     steps:
     - uses: actions/checkout@v4
 
-    - name: Setup .NET 6
-      uses: actions/setup-dotnet@v3
+    - name: Setup .NET
+      uses: actions/setup-dotnet@v4
       with:
-        dotnet-version: '6.0.x'
+        dotnet-version: |
+          6.0.x
+          8.0.x
       env:
         NUGET_AUTH_TOKEN: RequiredButNotUsed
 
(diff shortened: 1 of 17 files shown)


andystaples:
Superseded by https://github.com/microsoft/durabletask-mssql/pull/295

PR #321, opened 2026-07-29
Fix entity cleanup paging

andystaples:
## Summary

- Re-query the current entity cleanup page after deleting empty entities so shifted rows are not skipped.
- Advance to the next page only when the current page contains no deletable entities.
- Deduplicate orphaned-lock release requests when cleanup revisits a page.
- Add an integration regression test covering 150 empty entities across multiple pages.

## Validation

- `dotnet build DurableTask.SqlServer.sln --no-restore --nologo --verbosity minimal`
- The targeted regression test builds but cannot complete locally because SQL Server rejects dynamically created integration-test logins; the pre-existing `EntityQueueModes.SwitchingToSeparateQueuesPartitionsExistingWork` test fails with the same login error.

### DIFF
--- a/src/DurableTask.SqlServer/EntitySqlBackendQueries.cs
+++ b/src/DurableTask.SqlServer/EntitySqlBackendQueries.cs
@@ -92,6 +92,7 @@ namespace DurableTask.SqlServer
 
             int retrievedResults = 0;
             IEnumerable<OrchestrationState> allResults = Array.Empty<OrchestrationState>();
+            var lockReleaseRequests = new HashSet<(string EntityId, string LockOwner)>();
             var stopwatch = new Stopwatch();
             stopwatch.Start();
             do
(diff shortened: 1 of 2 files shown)


src/DurableTask.SqlServer/EntitySqlBackendQueries.cs
```diff
@@ -105,10 +106,8 @@ public async override Task<CleanEntityStorageResult> CleanEntityStorageAsync(Cle
                     CreatedTimeTo = DateTime.MaxValue,
                     FetchInput = true,
```
bachuv: Can `FetchInput` be false here?

andystaples: Yes. Cleanup reads entity metadata from custom status, so I changed `FetchInput` to `false` to avoid loading unused input payloads.

bachuv (approved):

Issue #317, opened 2026-06-10
Exception: Trying to pass a table-valued parameter with 13 column(s) where the corresponding user-defined table type requires 14 column(s).


CodeRedick:
We're seeing this error in our logs after updating to 1.6.0. Luckily it doesn't actually seem to be breaking anything but I don't know if there's any side effects we haven't noticed yet.

Stack trace is attached.

[SqlDurableTaskIssueStackTrace.json](https://github.com/user-attachments/files/28802776/SqlDurableTaskIssueStackTrace.json)

cgillum:
Hi @CodeRedick. The StackTrace.json file you shared doesn't include any error messages. Can you please share the error message?

CodeRedick:
The error message is in the title of the issue, but here's a raw stack trace that's probably better anyway:

```text
TaskOrchestrationDispatcher-2591660e646945419dd1b847e30fdc71-0: Unhandled exception with work item 'e399cb306cc6479f834e3296e85050b0': Microsoft.Data.SqlClient.SqlException (0x80131904): Trying to pass a table-valued parameter with 13 column(s) where the corresponding user-defined table type requires 14 column(s).
   at Microsoft.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at Microsoft.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   at Microsoft.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   at Microsoft.Data.SqlClient.SqlCommand.FinishExecuteReader(SqlDataReader ds, RunBehavior runBehavior, String resetOptionsString, Boolean isInternal, Boolean forDescribeParameterEncryption, Boolean shouldCacheForAlwaysEncrypted)
   at Microsoft.Data.SqlClient.SqlCommand.CompleteAsyncExecuteReader(Boolean isInternal, Boolean forDescribeParameterEncryption)
   at Microsoft.Data.SqlClient.SqlCommand.InternalEndExecuteNonQuery(IAsyncResult asyncResult, Boolean isInternal, String endMethod)
[...]

PR #323, opened 2026-07-31
Guard published TVP contracts against schema drift

andystaples:
## Summary
- add an integration regression test that submits frozen v1.6 TVP record shapes to the latest database schema
- document that published TVPs are immutable outside major/breaking releases and should be versioned instead of altered
- add repository-wide Copilot instructions covering scope, conventions, testing, SQL performance, and rolling-upgrade compatibility
- document shared-schema rolling-upgrade considerations for multitenant deployments

This guards the mixed-version deployment scenario reported in #317 without adding runtime metadata checks or hot-path overhead.

## Testing
- `dotnet build test\DurableTask.SqlServer.Tests\DurableTask.SqlServer.Tests.csproj --no-restore`
- `dotnet test test\DurableTask.SqlServer.Tests\DurableTask.SqlServer.Tests.csproj --no-build --filter FullyQualifiedName~DatabaseManagement --verbosity minimal`


src/DurableTask.SqlServer/Scripts/README.md
```diff
@@ -43,7 +43,17 @@ The existing `schema-x.y.z.sql` files generally should NOT be modified after the
 1. Create a new `schema-x.y.z.sql` script file with the new schema. The `x.y.z` numbers should match the new nuget package version that will be shipped with this new script.
 1. Copy/paste the generic warning comments from any existing `schema-x.y.z.sql` files into the new script file as appropriate.
 1. For adding new columns or indexes, use the appropriate `ALTER TABLE` statements as well as the `IF NOT EXISTS` syntax to avoid errors if the column or index already exists. These scripts must be safe to run multiple times.
-1. For making changes to custom types, add `DROP TYPE` statements in the `schema-x.y.z.sql` file and then update the existing `CREATE TYPE` statements in the `logic.sql` file to ensure that those types get recreated with the newest schema.
+1. Do not change an existing user-defined table type used as a table-valued parameter (TVP) in a minor or patch release. Published TVPs are immutable client/database contracts. Adding even a nullable column causes SQL Server to reject records sent by older clients before the stored procedure begins executing.
+
+### Table-valued parameter compatibility
```
cgillum, line 48: Non-blocking: It looks like this content is duplicated across copilot-instructions.md and here. Consider pointing the instructions file to this file explicitly rather than duplicating content to avoid drift.

test/DurableTask.SqlServer.Tests/Integration/DatabaseManagement.cs
```diff
@@ -469,6 +470,118 @@ public async Task SchemaUpgradeAddsTagsColumn()
...
+                SqlParameter parameter = command.Parameters.Add("@Rows", SqlDbType.Structured);
+                parameter.TypeName = $"dt.{typeName}";
+                parameter.Value = new[] { record };
+
+                long rowCount = (long)(await command.ExecuteScalarAsync())!;
+                Assert.Equal(1, rowCount);
+            }
+        }
+
+        static (string TypeName, SqlMetaData[] Columns)[] GetPublishedTableValuedParameterContracts()
```
cgillum: Should we update the name of this method to reflect that this is the 1.6.0 schema? I'm wondering what the procedure is for updating this as we make more changes in the future.

cgillum (commented):
Thanks for creating this PR. I do have one question about the maintenance expectations.

src/DurableTask.SqlServer/Scripts/README.md
andystaples: Agreed. I reduced the duplicated section in copilot-instructions.md to the essential invariant and linked to Scripts/README.md as the canonical source for the rationale and maintenance procedure. I kept the short rule in the instructions file because repository instructions are injected automatically, while linked files are not guaranteed to be loaded or followed by Copilot.

test/DurableTask.SqlServer.Tests/Integration/DatabaseManagement.cs
andystaples: Agreed. I renamed the test and helper to identify the frozen v1.6.0 baseline. The schema README now documents the maintenance procedure: never modify an existing baseline; when a new versioned TVP is published, add a separate version-specific baseline and retain earlier baselines while those clients remain supported.

cgillum (approved):

PR #330, opened 2026-09-13
fix issue 148 - sub-orchestrations cannot reuse existing instance ids

### LINKED ISSUE #148: Sub-orchestrations cannot reuse existing instance IDs
As of v1.1.0, the current implementation of sub-orchestrations doesn't support reusing instance IDs.

If you try to schedule an orchestration using an instance ID of some previously completed orchestration, you will see the following message in the logs:

> 2023-02-01T18:06:39.6352615-08:00 [DurableTask.SqlServer] XYZ: Discarding [ExecutionStarted#0]: Target is in the Completed state.

The issue is specific to the MSSQL backend. The same scenario works fine with the Azure Storage backend.

usemam:
Fixes #148

Terminal targets (`Completed`/`Failed`/`Terminated`) are now purged via the existing `PurgeInstanceStateByID` before the insert, mirroring `CreateInstance`'s overwrite path. Active instances are untouched.

### Tests

Re-enabled the disabled `RetryFailedSubOrchestration(true)` case; added `ReuseCompletedSubOrchestrationID` and `CannotRecreateRunningSubOrchestrationInstance`.

### DIFF
--- a/src/DurableTask.SqlServer/Scripts/logic.sql
+++ b/src/DurableTask.SqlServer/Scripts/logic.sql
@@ -817,6 +817,24 @@ BEGIN
         SET @IsContinueAsNew = 1
     END
 
+    -- Sub-orchestration instance ID reuse case: https://github.com/microsoft/durabletask-mssql/issues/148
+    -- Purge the existing instance data so that the "Create sub-orchestration instances" statement further below can overwrite it.
+    -- NOTE: This must happen before any new payloads or events are written below, otherwise the
+    --       incoming sub-orchestration state would be purged along with the existing state.
+    DECLARE @SubOrchestrationsToRecreate InstanceIDs
+    INSERT INTO @SubOrchestrationsToRecreate ([InstanceID])
+        SELECT DISTINCT E.[InstanceID]
+        FROM @NewOrchestrationEvents E
+            INNER JOIN Instances I ON I.[TaskHub] = @TaskHub AND I.[InstanceID] = E.[InstanceID]
+        WHERE E.[EventType] = 'ExecutionStarted'
+            AND E.[InstanceID] <> @InstanceID -- never touch the instance being checkpointed
+            AND I.[RuntimeStatus] IN ('Completed', 'Failed', 'Terminated')
+
+    IF EXISTS (SELECT 1 FROM @SubOrchestrationsToRecreate)
+    BEGIN
+        EXEC __SchemaNamePlaceholder__.PurgeInstanceStateByID @SubOrchestrationsToRecreate
+    END
+
     -- Custom status case #1: Setting the custom status for the first time
     IF @ExistingCustomStatusPayload IS NULL AND @CustomStatusPayload IS NOT NULL
     BEGIN
--- a/test/DurableTask.SqlServer.Tests/Integration/Orchestrations.cs
+++ b/test/DurableTask.SqlServer.Tests/Integration/Orchestrations.cs
@@ -654,7 +654,7 @@ namespace DurableTask.SqlServer.Tests.Integration
         }
 
         [Theory]
-        //[InlineData(true)] // BUG: https://github.com/microsoft/durabletask-mssql/issues/148
+        [InlineData(true)]
         [InlineData(false)]
         public async Task RetryFailedSubOrchestration(bool userSpecifiedInstanceId)
         {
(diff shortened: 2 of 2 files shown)


usemam:
@cgillum could please take a look? Thank you.

src/DurableTask.SqlServer/Scripts/logic.sql
```diff
@@ -817,6 +817,24 @@ BEGIN
...
+    -- Purge the existing instance data so that the "Create sub-orchestration instances" statement further below can overwrite it.
+    -- NOTE: This must happen before any new payloads or events are written below, otherwise the
+    --       incoming sub-orchestration state would be purged along with the existing state.
+    DECLARE @SubOrchestrationsToRecreate InstanceIDs
+    INSERT INTO @SubOrchestrationsToRecreate ([InstanceID])
+        SELECT DISTINCT E.[InstanceID]
+        FROM @NewOrchestrationEvents E
+            INNER JOIN Instances I ON I.[TaskHub] = @TaskHub AND I.[InstanceID] = E.[InstanceID]
+        WHERE E.[EventType] = 'ExecutionStarted'
+            AND E.[InstanceID] <> @InstanceID -- never touch the instance being checkpointed
```
andystaples: **[P1] Keep the terminal-state check stable through purge and recreation**

This SELECT does not retain locks under READ COMMITTED, and can read an older version under READ_COMMITTED_SNAPSHOT. `PurgeInstanceStateByID` then deletes solely by instance ID, without rechecking the execution or terminal status. Another parent checkpoint or `CreateInstance` call can therefore recreate the selected child before this checkpoint reaches the DELETE, and this checkpoint will purge the new Pending/Running execution.

I reproduced this with two connections running the actual PR SQL: A selected the Completed child and blocked entering purge; B recreated that child as Pending and committed; A resumed, deleted B's start event, and installed its own child. B's parent retained its committed child-scheduling history but was left waiting for a deleted execution.

Please serialize and revalidate the child execution/status through replacement, while preserving consistent lock ordering, and add a deterministic concurrent-reuse regression. `CannotReuseRunningSubOrchestrationInstanceID` only covers an instance that was already Running before this SELECT, so it does not exercise this race.

src/DurableTask.SqlServer/Scripts/logic.sql
```diff
@@ -817,6 +817,24 @@ BEGIN
...
+            AND I.[RuntimeStatus] IN ('Completed', 'Failed', 'Terminated')
+
+    IF EXISTS (SELECT 1 FROM @SubOrchestrationsToRecreate)
+    BEGIN
+        EXEC __SchemaNamePlaceholder__.PurgeInstanceStateByID @SubOrchestrationsToRecreate
```
andystaples: **[P1] Roll back the child purge when checkpoint deduplication fails**

This destructive purge happens before the parent History insert used to detect duplicate/split-brain checkpoints. If a parent checkpoint commits, its child completes, and the same checkpoint is replayed after a lost response or by a stale worker, this code purges the completed child before detecting the duplicate parent history.

The later History insert raises error 2627, but `_CheckpointOrchestration` has neither `XACT_ABORT ON` nor a TRY/CATCH rollback. With the default SQL session settings, execution continues to COMMIT, and `CompleteTaskOrchestrationWorkItemAsync` subsequently treats the unique-key error as a harmless duplicate.

I reproduced the rejected replay leaving the child Pending, with its history deleted, a fresh ExecutionStarted event queued, and `@@TRANCOUNT = 0`. The identical replay against the merge-base SQL preserves the child's Completed status and history. The rollback gap already exists, but this new purge turns it into deletion and re-execution of an already-completed child.

Please ensure checkpoint failure rolls back the entire transaction, or reject duplicates before destructive work, and add a regression that replays a checkpoint after its child has completed. Otherwise completed sub-orchestrations and their activities can execute again.

andystaples (changes requested):
Requesting changes for two P1 data-consistency issues in the new sub-orchestration instance-ID reuse path:

1. A concurrent checkpoint can purge another parent's newly active child execution.
2. A duplicate checkpoint can commit a purge and restart of a completed child even though checkpoint deduplication raises an error.

Both issues were reproduced against this PR's SQL in an isolated SQL Server 2022 database. For the second issue, the same replay against the merge-base SQL preserves the completed child's status and history. The inline comments include the failure sequences and requested changes.

